PR #4, opened 2025-12-30
Feat/basic spaceship domain resource

dm1tr1yvovk:
there is a lot of code changes, but they are mostly moving code in different place.
you may mostly focus on internal/provider/domain_resource.go and related acceptance tests

### DIFF
--- a/internal/provider/domain_resource.go
+++ b/internal/provider/domain_resource.go
@@ -144,6 +144,19 @@ func (d *domainResource) Delete(_ context.Context, req resource.DeleteRequest, r
 	// leaving infra in the same state
 }
 
+func (d *domainResource) ModifyPlan(ctx context.Context, req resource.ModifyPlanRequest, resp *resource.ModifyPlanResponse) {
+	// Handle destruction
+	if req.Plan.Raw.IsNull() {
+		resp.Diagnostics.AddWarning(
+			"Resource Destruction Considerations",
+			"Applying this resource destruction will only remove the resource from the Terraform state "+
+				"and will not call the deletion API due to nature of domain specifics. "+
+				"Your registered domain and its settings would remain intact",
+		)
+		return
+	}
+}
+
 func (d *domainResource) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
 	var plan, state domainResourceModel
 


internal/client/client.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"net/http"
+	"strings"
+	"time"
+)
+
+// Client provides a thin wrapper around Spaceship API
+type Client struct {
+	baseURL    string
```
voiddev: It is better to use types. Golang has `url.URL` from standard library `net/url`

internal/client/client.go
```diff
@@ -0,0 +1,32 @@
...
+	apiKey     string
+	apiSecret  string
+	httpClient *http.Client
+}
+
+// created a new Spaceship API Client
```
voiddev: Please revise source code docs, add more details

internal/client/client.go
```diff
@@ -0,0 +1,32 @@
...
+		apiSecret: apiSecret,
+		httpClient: &http.Client{
+			Timeout: 30 * time.Second,
+		},
+	}
+}
+
+func (c *Client) applyAuth(req *http.Request) {
+	req.Header.Set("X-API-Key", c.apiKey)
+	req.Header.Set("X-API-Secret", c.apiSecret)
```
voiddev: Prefix `X-API` can be a reused constant

internal/client/dns.go
```diff
@@ -0,0 +1,190 @@
...
+	Matching        *int         `json:"matching,omitempty"`
+	AssociationData string       `json:"associationData,omitempty"`
+	Group           *RecordGroup `json:"group,omitempty"`
+}
+
+type RecordGroup struct {
+	Type string `json:"type"`
+}
+
+type PortValue struct {
```
voiddev: It is much more stable to use only one type for typing port. And `int` is much more robust for port number. Right now an instance state is mixed, which has the next problems:
- compile time safety - fields can contradict each other, for example one field is null another one is defined, has different values at the same time
- from a design point of view, the instance state is represented by joining two fields together at the same time, which is confusing
- to cover all edge cases, you need a lot of code checking, casting, validation and syncing both fields
- memory inefficient

internal/client/client.go
```diff
@@ -0,0 +1,32 @@
...
+// Client provides a thin wrapper around Spaceship API
+type Client struct {
+	baseURL    string
+	apiKey     string
+	apiSecret  string
+	httpClient *http.Client
+}
+
+// created a new Spaceship API Client
+func NewClient(baseURL, apiKey, apiSecret string) *Client {
```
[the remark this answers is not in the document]
dm1tr1yvovk: no, this code is used. it is part of library, used outside of this package check provider.go file

PR #36, opened 2026-04-01
fix: filter out only custom dns records

### LINKED ISSUE #21: DNS resource should scope management to External API / Terraform DNS groups, not the whole zone
## Summary

The Terraform provider currently models DNS management as ownership of the full zone via `spaceship_dns_records`.

That does not fit how Spaceship itself structures DNS when first-party features are enabled. Spaceship has the concept of DNS groups / connection-owned records. Records created and managed by built-in features should stay under Spaceship's control and should not need to be declared in Terraform.

## Problem

When features such as URL Redirect are enabled in Spaceship, Spaceship creates DNS records inside a DNS group associated with that connection.

Example shape of such records:

- `A @ -> 15.197.162.184`
- `A * -> 15.197.162.184`

These are not user-authored Terraform records. They belong to a Spaceship-managed connection / DNS group.

At the same time, for external automation there is a DNS group named `External API Custom Group`.

The provider should not behave as if it owns and replaces every record in the zone across all groups.

## Expected behavior

The provider should be blind to DNS records that belong to Spaceship connection-managed groups.

Instead, Terraform-managed records should be scoped to something like:

1. The existing `External API Custom Group`, or
2. A dedicated provider-managed DNS group, or
3. A unique Terraform-managed group per resource/module instance.

## Why this matters

The current whole-zone contract creates conflicts with Spaceship-native features:

- URL Redirect and similar features create records that Terraform should not need to manage.
- Users are forced to manually mirror Spaceship-managed records into Terraform just to avoid drift or provider errors.
- Multiple unrelated Terraform modules cannot safely manage records for the same domain unless they all coordinate on one huge combined zone definition.

This is especially problematic for monorepos or separate infrastructure stacks that only need to manage a subset of records for the same domain.

## Suggested direction

A better model would be one of the following:

### Option 1

Make the provider operate only against `External API Custom Group` and ignore all other groups returned by the API.

### Option 2

Create and manage a provider-owned DNS group, for example a Terraform-specific group, and only reconcile records inside that group.

### Option 3

Allow the resource to declare a group identifier / group name so different Terraform modules can manage different groups independently for the same domain.

If multiple Terraform modules are expected to coexist, the group should ideally be unique per resource/module instance rather than global for the domain.

## Benefit

This would align the provider with Spaceship's own DNS grouping model and avoid collisions with built-in products and connections managed directly in the Spaceship UI.

dm1tr1yvovk:
## Description

filters only custom dns records

## Related Issues

fixes #21

## Checklist

- [x] Unit tests added or updated
- [x] `make test` passes
- [x] `golangci-lint run ./...` passes
- [x] `go build .` passes
- [x] `make docs` run if schema changed (and changes committed)

### DIFF
--- a/internal/provider/dns_records_helpers_test.go
+++ b/internal/provider/dns_records_helpers_test.go
@@ -76,6 +76,20 @@ func TestRecordValueSignature_CAAValueCaseInsensitive(t *testing.T) {
 	}
 }
 
+func TestRecordValueSignature_SvcParamsCaseInsensitive(t *testing.T) {
+	r1 := client.DNSRecord{Type: "HTTPS", Name: "@", SvcPriority: intPtr(1), TargetName: "target.com", SvcParams: "ALPN=H2", Port: client.NewStringPortValue("_443"), Scheme: "_https"}
+	r2 := client.DNSRecord{Type: "HTTPS", Name: "@", SvcPriority: intPtr(1), TargetName: "target.com", SvcParams: "alpn=h2", Port: client.NewStringPortValue("_443"), Scheme: "_https"}
+	if recordValueSignature(r1) != recordValueSignature(r2) {
+		t.Error("expected case-insensitive match for HTTPS SvcParams")
+	}
+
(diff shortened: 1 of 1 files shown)


kurok (changes requested):
Please fix as minimum
"Empty Group.Type silently dropped" - this is second issue.
Nice to have
additional tests cases and update documentation

PR #50, opened 2026-04-10
fix: add validation for a record

dm1tr1yvovk:


kurok (changes requested):
Not ready to merge -- the int64-to-uint16 silent truncation  is a correctness bug that can produce wrong behavior.
   The discarded regex error  is a robustness concern for edge cases. Fix those two, and the rest are improvements that   
  could land in a follow-up.

internal/provider/records/srv_record_validator.go
```diff
@@ -108,17 +97,24 @@ func (v *srvRecordValidator) ValidateObject(ctx context.Context, req validator.O
...
-	// Validate port_number: required, 1-65535
 	portNumberAttr, _ := attrs["port_number"].(types.Int64)
 	if portNumberAttr.IsNull() || portNumberAttr.IsUnknown() {
 		resp.Diagnostics.AddAttributeError(
 			req.Path.AtName("port_number"),
 			"Missing Required Field",
 			"The 'port_number' field is required for SRV records.",
 		)
+	} else {
+		rec.Port = uint16(portNumberAttr.ValueInt64())
```
kurok: **Must Fix — silent int64-to-uint16 truncation**

If Terraform passes `port_number = 65536`, `ValueInt64()` returns 65536, the cast `uint16(65536)` silently wraps to `0`, and the user sees the confusing error "must be between 1 and 65535, got 0". Worse, `port_number = 131073` wraps to `1` and *passes* validation with a wrong value.

The `client/records` package is designed to be usable outside Terraform where schema-level `int64validator.Between()` guards don't exist.

Suggestion: either validate the int64 range before casting:
```go
portVal := portNumberAttr.ValueInt64()
if portVal < 1 || portVal > 65535 {
    // report error
} else {
    rec.Port = uint16(portVal)
}
```
Or change `SRVRecord.Port` to `int` (matching `TTL`) so `ValidatePort()` handles the full range itself.

kurok (commented):
## Code Review

The architectural direction is solid — separating pure validation logic from Terraform plumbing into `client/records` makes it reusable for any Go SDK consumer and cleans up the test structure. Two correctness issues need fixing before merge; the rest are suggestions.

See inline comments below, grouped by severity.

internal/client/records/common.go
```diff
@@ -0,0 +1,28 @@
...
+var hostnamePattern = regexp2.MustCompile(`^(?!\.)(@|\*|([_*]\.)?(?:(?!-)(?=[^\.]*[^\W_])[\w-]{1,63}(?<!-)($|\.)){1,127}(?<!\.))$`, regexp2.None)
+
+// ValidateName checks that a record name is 1-253 chars and matches the hostname pattern.
+func ValidateName(name string) error {
+	if len(name) < 1 || len(name) > 253 {
+		return fmt.Errorf("must be between 1 and 253 characters, got %d", len(name))
+	}
+	if matched, _ := hostnamePattern.MatchString(name); !matched {
+		return fmt.Errorf("must be a valid hostname (e.g. 'myhost' or '@'), got %q", name)
+	}
```
kurok, line 22: **Must Fix — discarded `regexp2` error**

`regexp2.MatchString` returns `(bool, error)`. The `regexp2` engine can return errors (e.g. backtracking limits on adversarial input). Discarding the error means a regex engine failure silently rejects valid hostnames.

Same pattern appears in `srv.go` line 69 where `hostnamePattern.MatchString` is also called.

```go
matched, err := hostnamePattern.MatchString(name)
if err != nil {
    return fmt.Errorf("hostname validation failed: %w", err)
}
if !matched {
    return fmt.Errorf("must be a valid hostname ...")
}
```

internal/client/records/a.go
```diff
@@ -0,0 +1,50 @@
+package records
+
+import (
+	"fmt"
+	"net"
+	"strings"
+)
+
+type ARecord struct {
+	Type    string
```
kurok: **Should Fix — `Type` field is dead code**

`Type` is never set, read, or validated by any code in this PR. Same applies to `SRVRecord.Type` in `srv.go`. If these exist for future serialization, add a comment explaining why; otherwise remove until needed.

internal/provider/records/a_record_validator.go
```diff
@@ -0,0 +1,63 @@
...
+		rec.Address = addressAttr.ValueString()
+		if err := rec.ValidateAddress(); err != nil {
+			resp.Diagnostics.AddAttributeError(
+				req.Path.AtName("address"),
+				"Invalid Address Value",
+				err.Error(),
+			)
+		}
+	}
+}
```
kurok, line 65: **Should Fix — validator only checks `address`, skips Name/TTL**

The `client/records.ARecord.Validate()` method runs all three checks (address, name, TTL), but this Terraform validator only validates `address`. If name/TTL are handled at the schema-attribute level, this is fine — but it's inconsistent with the `Validate()` contract.

Consider either:
- Adding a brief comment: `// Name and TTL are validated by schema-level attribute validators`
- Or calling `rec.Validate()` and mapping all returned errors to diagnostics

internal/client/records/a.go
```diff
@@ -0,0 +1,50 @@
...
+	Address string
+	Name    string
+	TTL     int
+}
+
+// ValidateAddress checks that the address is a valid IPv4 address up to 15 characters.
+func (r *ARecord) ValidateAddress() error {
+	if len(r.Address) > 15 {
```
kurok: **Nit — redundant length check**

Any string longer than 15 characters cannot be a valid IPv4 address, so `net.ParseIP` on line 21 would catch it. This produces a less informative error ("must be at most 15 characters") than the generic path ("must be a valid IPv4 address"). Consider removing and letting `ParseIP` handle all invalid inputs uniformly.

internal/client/records/a_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+func TestARecord_ValidateTTL(t *testing.T) {
+	tests := []struct {
+		name    string
+		ttl     int
+		wantErr bool
+	}{
+		{"valid", 3600, false},
+		{"min valid", 60, false},
+		{"max valid", 3600, false},
```
kurok, line 87: **Nit — `"max valid"` duplicates `"valid"`**

Both use `3600`. If the actual max TTL is ever changed to something higher, this test won't catch an off-by-one at the upper bound. Consider testing a value different from `"valid"`, or testing just below the upper limit.

Same issue in `srv_test.go`.

internal/client/records/srv.go
```diff
@@ -0,0 +1,84 @@
...
+	Port     uint16
+	Target   string
+	Name     string
+	TTL      int
+}
+
+func (r *SRVRecord) ValidateService() error {
+	if len(r.Service) < 2 || len(r.Service) > 63 {
+		return fmt.Errorf("must be between 2 and 63 characters, got %d", len(r.Service))
+	}
```
kurok, line 25: **Nit — missing doc comments on exported methods**

`ValidateService`, `ValidateProtocol`, `ValidatePort`, `ValidateTarget` (and the delegating `ValidateName`/`ValidateTTL`) are exported but lack doc comments. Per Go convention, exported methods should have doc comments starting with the method name. The shared helpers in `common.go` already follow this pattern.

internal/client/records/a.go
dm1tr1yvovk: okay, just wanted to have 1 to 1 parity with api documentation, but you are right that it is redundant
deleted this check.

internal/provider/records/srv_record_validator.go
dm1tr1yvovk: agree

kurok (approved):
Duplicate test value "max valid" (a_test.go) but this not an big issue

kurok (approved):

kurok (approved):

PR #58, opened 2026-04-15
fix: add validation for aaaa record

dm1tr1yvovk:


internal/provider/records/aaaa_record_validator.go
```diff
@@ -0,0 +1,65 @@
...
+	if addressAttr.IsNull() || addressAttr.IsUnknown() {
+		resp.Diagnostics.AddAttributeError(
+			req.Path.AtName("address"),
+			"Missing Required Field",
+			"The 'address' field is required for AAAA records.",
+		)
+	} else {
+		rec.Address = addressAttr.ValueString()
+		if err := rec.ValidateAddress(); err != nil {
+			resp.Diagnostics.AddAttributeError(
```
kurok, line 64: Bug: the `ok` value from the type assertion is discarded. If the assertion fails (key missing or wrong type), `addressAttr` is the zero value of `types.String` — its `IsNull()` returns `true`, producing a misleading "Missing Required Field" error instead of "Invalid Field Type".

This is the same pattern PR #57 just fixed in the A and SRV validators. Should match the established pattern:

```suggestion
	addressAttr, ok := attrs["address"].(types.String)
	if !ok {
		resp.Diagnostics.AddAttributeError(
			req.Path.AtName("address"),
			"Invalid Field Type",
			"The 'address' field must be a string for AAAA records.",
		)
	} else if addressAttr.IsNull() || addressAttr.IsUnknown() {
```

kurok (changes requested):
## Review: AAAA record validation

Overall solid work — the validation logic, test coverage, and housekeeping changes are all good. One issue needs fixing before merge.

### Bug: missing type assertion `ok` check in `aaaa_record_validator.go`

This is ironic since PR #57 was specifically about adding `ok` checks to the A and SRV validators to prevent this exact class of bug. The new AAAA validator introduces the same pattern that was just fixed.

When the type assertion fails, `addressAttr` becomes the zero value of `types.String`, whose `IsNull()` returns `true` (since `ValueStateNull = 0`). The user gets a misleading "Missing Required Field" error instead of the correct "Invalid Field Type" error. See inline suggestion for the fix.

### What looks good

- **Validation logic** (`aaaa.go`): `net.ParseIP` + `ip.To4() != nil` correctly rejects both plain IPv4 and IPv4-mapped IPv6, matching API behavior. The 39-char length cap is well-documented.
- **Unit tests** (`aaaa_test.go`): Well-structured table-driven tests with good edge case coverage (empty, too-long, IPv4, IPv4-mapped, loopback, unspecified).
- **Acceptance tests**: Thorough happy-path and error-path coverage. `aaaaHost()` helper prevents test collision — good pattern.
- **Housekeeping**: Struct alignment fix, removal of stale TODO comment, and `Fail;ed` → `Failed` typo fix are all welcome.

kurok (dismissed):

internal/provider/records/aaaa_record_validator.go
dm1tr1yvovk: In practice this can only fail on a schema/validator mismatch — the framework guarantees `attrs["address"]` is a `types.String` for a `schema.StringAttribute`, so end users can't trigger the misleading diagnostic path. That said, applied for consistency with #57 and so the diagnostic stays honest if the schema ever drifts.

kurok:
## Review: AAAA record validation

The type-assertion `ok` check bug from the previous review has been properly addressed — the latest commit handles the `!ok` path with a distinct "Invalid Field Type" diagnostic, consistent with the A and SRV validators from PR #57. Nicely done.

Two major issues remain:

### 1. Makefile regex `Test[^Acc]` is a no-op change — still silently skips AAAA unit tests

`[^Acc]` is a **character class**, not a string negation. In regex, `[^Acc]` = `[^Ac]` = "any character that is NOT `A` or `c`". It behaves identically to the original `Test[^A]` for standard Go test names (which always start uppercase after `Test`).

This means `make test` and `make test-cover` still skip **all** tests starting with `TestA*`, including the new `TestAAAARecord_*` unit tests in `aaaa_test.go`. You can verify:

```python
import re
re.search(r'Test[^Acc]', 'TestAAAARecord_Validate')  # → None (skipped!)
re.search(r'Test[^Acc]', 'TestAccDNSRecords_basic')   # → None (skipped, intended)
```

The proper fix: acceptance tests already guard themselves with `t.Skip` via `testAccPreCheck(t)`, so the Makefile can simply use `go test ./...` and let the skip guards do their job. Alternatively, `go test -run '^Test[^(Acc)]'` won't work either (POSIX character classes don't support grouping). The `t.Skip` approach is the idiomatic Go pattern and the only robust solution here.

### 2. Missing unit tests for `aaaa_record_validator.go` — 0% coverage on new provider code

The existing A and SRV validators both have unit test files:
- `internal/provider/records/a_record_validator_test.go`
- `internal/provider/records/srv_record_validator_test.go`

The new AAAA validator has **no unit test counterpart** — only acceptance tests (which require live API credentials and don't run in CI coverage). Codecov reports 0% patch coverage on `aaaa_record_validator.go` (37 lines missing). Adding `internal/provider/records/aaaa_record_validator_test.go` with table-driven cases for the `ValidateObject` method (valid IPv6, IPv4 rejection, missing address, wrong type) would bring this in line with the established pattern and provide CI-visible coverage.

Makefile
```diff
@@ -2,11 +2,11 @@ SHELL := /bin/bash
 
 .PHONY: test
 test:
-	go test -run 'Test[^A]' ./...
+	go test -run 'Test[^Acc]' ./...
```
kurok, line 5: this is not change anything

internal/provider/records/aaaa_record_validator.go
kurok, line 1: Missing unit tests for the AAAA validator

kurok (changes requested):

internal/provider/records/aaaa_record_validator.go
dm1tr1yvovk: they are done in aaaa_test.go on another level
there is no need to dublicate same tests in provider validator
there are acceptance tests for this.
this is my idea

Makefile
dm1tr1yvovk: yep, true

kurok (approved):

dm1tr1yvovk:
@kurok why to merge if there is a bug in test selection that was not fixed?

dm1tr1yvovk:
test selection fixed in #60

PR #61, opened 2026-04-16
feat: add validators and tests for alias record

dm1tr1yvovk:


kurok:
## Code Review Findings

No must-fix issues found. One **should fix** item below.

---

### Should Fix

**`internal/client/records/alias_test.go` — "too long" test cases use null bytes, not a long hostname**

Both `TestALIASRecord_ValidateAliasName` and `TestALIASRecord_ValidateName` have a `"too long"` case using `string(make([]byte, 254))`. This produces 254 null bytes (`\x00`), which a hostname validator rejects on **invalid characters** before ever reaching the length check. If the length check were removed, these tests would still pass — so they don't actually cover what they claim.

**Fix:** use a real 254-character hostname string:

```go
// before
{"too long", string(make([]byte, 254)), true},

// after
{"too long", strings.Repeat("a", 254), true},
```

Apply to both `ValidateAliasName` and `ValidateName` test tables.

kurok (approved):

PR #62, opened 2026-04-20
feat: add validators and tests for caa record

dm1tr1yvovk:


.github/workflows/ci.yml
```diff
@@ -31,6 +31,11 @@ jobs:
               with:
                   go-version-file: go.mod
 
+            - name: Set up Terraform
```
kurok: **Should Fix** — the `lint` job only runs `golangci-lint`, which doesn't need the Terraform binary. The same step is correctly gated with `if: ${{ steps.acc_secrets.outputs.has_secrets == 'true' }}` in the acceptance job below, but here it's unconditional. Either drop it from the lint job, or gate it the same way — saves ~10-15 s per CI run with no consumer.

```suggestion
```

(remove the four-line block entirely, or wrap in `if: false` if you expect a near-term need).

internal/provider/dns_records_resource.go
```diff
@@ -1030,7 +1031,6 @@ func recordValueSignature(record client.DNSRecord) string {
 	case "CNAME":
 		write(strings.ToLower(record.CName))
 	case "HTTPS":
-		// TODO looks like a bug  when svc prirotity is present in https record
 		write(intToString(record.SvcPriority), strings.ToLower(record.TargetName), strings.ToLower(record.SvcParams), portValueSignature(record.Port), strings.ToLower(record.Scheme))
```
kurok, line 1034: Nit — the TODO comment above this line was silently removed in this PR:

```go
// TODO looks like a bug  when svc prirotity is present in https record
```

The signature-building logic is unchanged. If the concern was investigated and ruled out, say so in the PR body (`TODO verified — HTTPS signature includes SvcPriority correctly`). If it was just visual cleanup, please open an issue first so the concern isn't lost — it's the kind of note that could matter next time HTTPS records show diff drift.

internal/provider/caa_record_validator_acc_test.go
```diff
@@ -0,0 +1,649 @@
...
+		}
+	}
+
+	// Clean up the pre-seeded record after the test.
+	t.Cleanup(func() {
+		testClient, err := testAccClient()
+		if err != nil {
+			return
+		}
+		_ = testClient.DeleteDNSRecords(context.Background(), domain, []client.DNSRecord{preExistingRecord})
```
kurok: **Should Fix** — both the `testAccClient()` init error and the `DeleteDNSRecords` error are dropped silently. If cleanup fails on this seeded record, the next run on the same `SPACESHIP_TEST_DOMAIN` collides on `tfacc-caa-import` and prints a confusing setup-time error that's hard to trace back to this test.

```suggestion
	t.Cleanup(func() {
		testClient, err := testAccClient()
		if err != nil {
			t.Logf("CAA cleanup: skipping — could not create test client: %v", err)
			return
		}
		if err := testClient.DeleteDNSRecords(context.Background(), domain, []client.DNSRecord{preExistingRecord}); err != nil {
			t.Logf("CAA cleanup: failed to delete seeded record %q: %v", host, err)
		}
	})
```

internal/client/records/caa.go
```diff
@@ -0,0 +1,79 @@
...
+	default:
+		return fmt.Errorf("must be one of 'issue', 'issuewild', 'iodef', got %q", r.Tag)
+	}
+}
+
+// ValidateValue checks that the value is 1-256 ASCII-printable characters,
+// matching the API's pattern ^[ -~]+$ and maxLength 256.
+func (r *CAARecord) ValidateValue() error {
+	if len(r.Value) < 1 || len(r.Value) > 256 {
+		return fmt.Errorf("must be between 1 and 256 characters, got %d", len(r.Value))
```
kurok, line 45: Nit — `len(r.Value)` measures bytes, but the error message says "characters". For ASCII-only input (what the regex two lines down enforces) bytes == runes, so the check is correct. The divergence only shows up if the user passes non-ASCII: they'd see e.g. `must be between 1 and 256 characters, got 768` for a 256-rune Chinese string.

Moot in practice because the regex rejects non-ASCII anyway. Lowest-friction fix is to tighten the comment to "bytes" consistently:

```suggestion
	if len(r.Value) < 1 || len(r.Value) > 256 {
		return fmt.Errorf("must be between 1 and 256 bytes, got %d", len(r.Value))
	}
```

internal/provider/records/caa_record_validator.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/types"
+
+	clientrecords "terraform-provider-spaceship/internal/client/records"
+)
+
+type caaRecordValidator struct{}
+
+var _ validator.Object = &caaRecordValidator{}
+
+func CAAValidator() validator.Object {
```
kurok, line 16: Nit — `CAAValidator` is exported so by Go style it should have a `// CAAValidator returns ...` doc comment. Flagging for the record only: `ALIASValidator`, `AAAAValidator`, `AValidator`, `SRVValidator` all omit this too, so fixing here in isolation creates asymmetry. Worth a separate follow-up PR that adds doc comments to all five constructors as a batch.

kurok:
## Review summary

I did a full code review. `gofmt -d`, `go vet`, and `make test` all clean against the PR head.

### Findings (by severity)

- **Must Fix:** none.
- **Should Fix:** 2 — see inline comments for details:
  1. Unconditional `Set up Terraform` in the `lint` job of `.github/workflows/ci.yml:34` (lint doesn't need Terraform; wastes ~10-15 s per run).
  2. `t.Cleanup` in `internal/provider/caa_record_validator_acc_test.go:267` silently drops both the test-client init error and the delete-records error. If cleanup fails the next run collides on `tfacc-caa-import` — add `t.Logf` on both paths.
- **Nit:** 3 — see inline comments for suggestion blocks:
  1. `internal/client/records/caa.go:45` — byte length vs "characters" wording (moot in practice since the regex rejects non-ASCII anyway).
  2. `internal/provider/records/caa_record_validator.go:16` — exported `CAAValidator()` lacks a doc comment (consistent with A/AAAA/ALIAS/SRV, so track as a codebase-wide cleanup).
  3. `internal/provider/dns_records_resource.go:1034` — the `// TODO looks like a bug when svc priority is present in https record` comment is removed without code change. If the concern was investigated, say so in the PR body; otherwise open an issue so the note isn't lost.

### Retraction

I posted an earlier inline comment on line 213 of `dns_records_resource.go` claiming `AAAAValidator()` and `ALIASValidator()` registrations were silent scope creep. **That was wrong** — I was diffing against a stale local `upstream/master` that pre-dated #57–#61. The real PR base (`fe254d4`, release 0.1.0) already has both registered, and the only net change at line 213 is the expected `records.CAAValidator()` addition. That comment is deleted.

Similarly, I had initially planned a Nit on a typo fix (`Fail;ed` → `Failed`) in the same file, but that fix already landed via #57 — not part of this PR. Disregard if I mentioned it anywhere.

### Positives

- **Test breadth is excellent** — 12 acceptance tests covering CRUD, 4 negative-validation variants, import-pre-existing, flag=128, value update (delete+upsert), case-insensitive round-trip, same-host multi-tag, mailto, and the `;` RFC 8659 "no CA may issue" form. `plancheck.ExpectEmptyPlan()` on re-apply is the right convergence signal.
- **Idiomatic client-layer validator** — package-level `regexp.MustCompile`, `var errs []error` (nil slice), closures-slice pattern in `Validate()` matching `alias.go`.
- **Good CLAUDE.md addition** — the three-file validator contract is the right doc to prevent silent registration omissions.
- **Clean dep bump** — all tagged releases, no pseudo-versions.

PR #64, opened 2026-04-21
feat: add validation with tests for cname record

dm1tr1yvovk:


kurok (commented):
## Code Review Summary

Adds a CNAME record validator following the established three-file pattern (client struct + tests, provider `validator.Object`, schema registration) plus acceptance tests. The implementation is a faithful copy of the ALIAS validator, adjusted for the `cname` attribute.

Base SHA diffed: `56d9c247` → `a02f31a4` (5 files, +593 LOC).

### Must Fix
_(none)_

### Should Fix
_(none)_

### Nits

- **`internal/client/records/cname.go:36-50`** — `Validate()` and `ValidateName()`/`ValidateTTL()` on `CNAMERecord` are exported but never called from production code (the provider validator only invokes `ValidateCName()`). This mirrors the ALIAS pattern, but it leaves three methods that exist solely for their own unit tests.

- **`internal/client/records/cname_test.go:16-21`** — `TestCNAMERecord_Validate_ValidRecord` only exercises the happy path. Consider adding a case where multiple fields are invalid to verify the aggregator collects all errors. Same gap exists in `alias_test.go`, so this is an ecosystem observation rather than CNAME-specific.

- **`internal/provider/records/cname_record_validator.go:55-60`** — `cnameAttr.IsUnknown()` is grouped with `IsNull()` and both produce a "Missing Required Field" diagnostic. An unknown value (e.g. an unresolved interpolation at plan time) isn't missing; the framework convention is typically to skip validation when the value is unknown. Matches all sibling validators, so it's a pre-existing concern.

- **`internal/provider/cname_record_validator_acc_test.go:16-24`** — The `cnameHost` comment says the suffix keeps "parallel/repeat runs" from colliding, but the output is deterministic per suffix, so it protects against *different* tests colliding (e.g. `basic` vs `apex`), not parallel runs of the same test.

### Automated Checks

- `gofmt -l` on new files — clean
- `go vet ./...` — clean
- `golangci-lint run ./...` — `0 issues.`
- `make test` — pass

**Bottom line:** LGTM. Clean, faithful mirror of the ALIAS validator and its acceptance tests. No must-fix or should-fix items.

kurok (approved):

PR #65, opened 2026-04-21
feat: add validation and tests for https record

dm1tr1yvovk:


kurok (commented):
## Code Review Summary

Adds HTTPS (RFC 9460) record validation: client-level `HTTPSRecord` struct with field-level validators, provider `validator.Object`, schema registration, and a broad acceptance test set (ServiceMode, AliasMode, mode exchange, multi-priority, wildcard port, apex name, TTL-only update, coexistence with A, config-removal).

Base SHA diffed: `56d9c247` → `f63086c1` (5 files, +1473 LOC).

### Must Fix

- **`internal/client/records/https.go:8`** — `httpsPortPattern` has a logic bug that rejects many valid ports in 60000-65535. The first alternation `6[0-5]{2}[0-3][0-5]` is clearly intended to cover 60000-65535 but actually matches only `6XYZW` where `X,Y ∈ [0-5]`, `Z ∈ [0-3]`, `W ∈ [0-5]`. A sweep of `60000..65535` shows **4672 of 5536 valid ports are rejected**: `_60009`, `_60099`, `_60999`, `_64999`, `_62345`, `_63456`, etc. A user who sets `port = "_64999"` will get `"must be '*' or '_N' where N is 1-65535, got '_64999'"` despite 64999 being a valid port. The schema-level `portStringPattern` (`^(\*|_[0-9]{1,5})$`) accepts these values, so only this Object validator blocks them.

  Recommended: replace the regex with `strconv.Atoi` on the `_`-stripped suffix and range-check 1-65535. That's easier to audit and avoids regex golf for integer bounds. A correct regex would be something like `^(\*|_([1-9]|[1-9][0-9]{1,3}|[1-5][0-9]{4}|6[0-4][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5]))$`.

### Should Fix

- **`internal/client/records/https_test.go:132-161`** — `TestHTTPSRecord_ValidatePort` has no coverage for the 60000-65535 band, which is exactly where the regex above goes wrong. Any of `_60009`, `_64999`, `_65530` as positive cases would have caught the bug. Add boundary cases around 60000 and 65535.

- **`internal/provider/records/https_record_validator.go:48-70`** — `svc_priority`'s 0-65535 range is already enforced at the schema layer by `int64validator.Between(0, 65535)` in `dns_records_resource.go:290`. The schema-level validator runs before the Object validator, so `ValidateSvcPriority()` at this layer is dead work (the client-side method and its unit tests still have value). Either drop the call here, or conversely drop the schema-level `Between` if the goal is to centralize in the Object validator — but don't duplicate.

### Nits

- **`internal/provider/records/https_record_validator.go:114-124, 126-136`** — `port` and `scheme` silently skip validation when the type-assertion `ok` is false; `svc_priority`/`target_name`/`svc_params` emit an "Invalid Field Type" diagnostic in the same situation. Minor inconsistency in how wrong-type attributes are reported. In practice the schema prevents the mismatch, so cosmetic.

- **`internal/client/records/https.go:55-69`** — The length pre-check `len(r.Port) < 2 || len(r.Port) > 6` is redundant with the anchored regex (which also bounds length). Not worth a change if the regex stays; worth removing if the validator moves to `strconv.Atoi`.

- **`internal/client/records/https.go:97-114`** — `HTTPSRecord.Validate()` and the `ValidateName`/`ValidateTTL` methods are never called from production code (`name` and `ttl` have schema-level validators). Mirrors the established ALIAS/CNAME pattern — not PR-specific, just noting.

### Automated Checks

- `gofmt -l` on new files — clean
- `go vet ./...` — clean
- `golangci-lint run ./...` — `0 issues.`
- `make test` — pass

**Bottom line:** Structure and test coverage are solid, but the port regex is a genuine bug that will false-reject a large chunk of valid configurations. Fix that and add a regression test; the other items are small improvements.

kurok (changes requested):
Requesting changes — see findings above. The port regex bug (**4672 of 5536 valid ports in 60000-65535 rejected**) is blocking.

internal/provider/records/https_record_validator.go
```diff
@@ -0,0 +1,151 @@
...
+			rec.Scheme = schemeAttr.ValueString()
+		}
+		if err := rec.ValidateScheme(); err != nil {
+			resp.Diagnostics.AddAttributeError(
+				req.Path.AtName("scheme"),
+				"Invalid Scheme Value",
+				err.Error(),
+			)
+		}
+	}
```
kurok: Nit: `ValidateScheme()` runs even when `schemeAttr.IsUnknown()`. With a known `port` (e.g. `_443`) and an unknown `scheme` (e.g. bound to another resource's computed attribute), this emits a phantom *"is required when port is set"* at plan time. Skip validation when the value isn't known yet:

```suggestion
	} else if !schemeAttr.IsUnknown() {
		if !schemeAttr.IsNull() {
			rec.Scheme = schemeAttr.ValueString()
		}
		if err := rec.ValidateScheme(); err != nil {
			resp.Diagnostics.AddAttributeError(
				req.Path.AtName("scheme"),
				"Invalid Scheme Value",
				err.Error(),
			)
		}
	}
```

internal/client/records/https.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+	if r.Port[0] != '_' {
+		return fmt.Errorf("must be '*' or '_N' where N is 1-65535, got %q", r.Port)
+	}
+	suffix := r.Port[1:]
+	for _, c := range suffix {
+		if c < '0' || c > '9' {
+			return fmt.Errorf("must be '*' or '_N' where N is 1-65535, got %q", r.Port)
+		}
+	}
```
kurok, line 65: Nit: the manual digit scan is defense-in-depth — `strconv.Atoi` below already rejects empty, non-digit, and overflow. The one extra case it still blocks is a signed suffix like `_+5`/`_-5` (Atoi accepts a leading sign). Since the schema-level `portStringPattern = ^(\*|_[0-9]{1,5})$` already rules signs out, the loop is belt-and-suspenders. Either drop it, or pin the intent with a one-liner:

```suggestion
	suffix := r.Port[1:]
	// reject leading '+' or '-' that strconv.Atoi would otherwise accept
	for _, c := range suffix {
		if c < '0' || c > '9' {
			return fmt.Errorf("must be '*' or '_N' where N is 1-65535, got %q", r.Port)
		}
	}
```

kurok (commented):
## Follow-up Review

Previous Must/Should/Nit findings are all addressed:
- port regex bug replaced with `strconv.Atoi` + range check
- 60000-65535 boundary cases added to `TestHTTPSRecord_ValidatePort`
- `TestAccDNSRecords_httpsRecordPortInHighRange` pins the fix end-to-end (`_64999`)
- duplicate schema-layer `int64validator.Between(0, 65535)` on `svc_priority` removed
- `port`/`scheme` wrong-type diagnostics now match the other fields

Base `79c68561` → head `240af7f` (5 files, +1544 LOC).

### Automated Checks
- `gofmt -l` — clean
- `go vet ./...` — clean
- `golangci-lint run ./...` — 0 issues
- `make test` — pass

### Findings

**Must Fix:** none.

**Should Fix:** none.

**Nits** (non-blocking, inline suggestions below):
1. `ValidateScheme` called unconditionally when `scheme` is `IsUnknown` → phantom "required when port is set" if port is known and scheme is computed.
2. Manual digit-loop in `ValidatePort` is defense-in-depth against `_+N`/`_-N` that `strconv.Atoi` would accept; worth either a comment or removal (schema `portStringPattern` already blocks signs).
3. Optional: a `_+5` negative case would pin down the digit-loop's intent.

**Bottom line:** approvable as-is. Nits are polish.

internal/client/records/https_test.go
```diff
@@ -0,0 +1,246 @@
...
+		{"valid _60009 high range", "_60009", false},
+		{"valid _60999 high range", "_60999", false},
+		{"valid _64999 high range", "_64999", false},
+		{"valid _65530 near upper boundary", "_65530", false},
+		{"invalid no underscore", "443", true},
+		{"invalid zero", "_0", true},
+		{"invalid too high", "_65536", true},
+		{"invalid non-numeric", "_abc", true},
+		{"invalid only underscore", "_", true},
+		{"invalid too long", "_123456", true},
```
kurok, line 153: Nit (optional): if the digit-loop in `ValidatePort` stays as defense-in-depth, a single signed-suffix case pins its intent and documents why it's not redundant with `strconv.Atoi`:

```suggestion
		{"invalid too long", "_123456", true},
		{"invalid signed suffix", "_+5", true},
```

kurok (dismissed):

PR #66, opened 2026-04-22
feat: add validation for mx record

dm1tr1yvovk:
fix: now srv record target rejects `@` and `*` as input


kurok (dismissed):
LGTM. Two nits, non-blocking:

1. **Scope**: SRV `target` now rejects `@`/`*` (`internal/client/records/srv.go:67-72`). Correct change (API 422s these), but it's a behavior change bundled in an MX-titled PR — worth a line in the description so release-please tags it in the changelog, or split into its own commit.
2. **Duplicate validation**: `preference` has `int64validator.Between(0, 65535)` at `dns_records_resource.go:309-311` and `rec.ValidatePreference()` at `mx_record_validator.go:87`. Same pattern #65 dropped for `svc_priority` — keep one layer for consistency.

No major issues. Automated checks clean (`gofmt`/`vet`/`golangci-lint`/`make test`).

PR #69, opened 2026-04-23
test: refactoring and shortening

dm1tr1yvovk:
based on feedback in private from @kurok 

items that I consider worth implement, some of them were considered as minor and not worth fixing for now.

### DIFF
--- a/internal/client/records/alias_test.go
+++ b/internal/client/records/alias_test.go
@@ -53,6 +53,3 @@ func TestALIASRecord_ValidateAliasName(t *testing.T) {
 		})
 	}
 }
-
-// Shared hostname (Name) and TTL edge cases live in common_test.go.
-// Per-type ValidateName / ValidateTTL wiring is covered by TestALIASRecord_Validate_ValidRecord.
--- a/internal/client/records/alias_test.go
+++ b/internal/client/records/alias_test.go
@@ -21,19 +21,22 @@ func TestALIASRecord_Validate_ValidRecord(t *testing.T) {
 }
 
 func TestALIASRecord_ValidateAliasName(t *testing.T) {
+	// wantErrContains pins the rejection branch for @/* so the test fails
+	// loudly if rejection ever fires from ValidateName instead of the pre-check.
 	tests := []struct {
-		name      string
-		aliasName string
-		wantErr   bool
+		name            string
+		aliasName       string
+		wantErr         bool
+		wantErrContains string
 	}{
-		{"valid hostname", "other.example.com", false},
-		{"valid subdomain", "sub.domain", false},
-		{"valid single label", "myhost", false},
-		{"apex rejected", "@", true},
-		{"wildcard rejected", "*", true},
-		{"empty", "", true},
-		{"too long", strings.Repeat("a", 254), true},
-		{"starts with dot", ".invalid", true},
+		{"valid hostname", "other.example.com", false, ""},
+		{"valid subdomain", "sub.domain", false, ""},
+		{"valid single label", "myhost", false, ""},
+		{"apex rejected", "@", true, "domain name"},
+		{"wildcard rejected", "*", true, "domain name"},
+		{"empty", "", true, ""},
+		{"too long", strings.Repeat("a", 254), true, ""},
+		{"starts with dot", ".invalid", true, ""},
 	}
 
 	for _, tt := range tests {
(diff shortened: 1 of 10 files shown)


internal/client/records/common_test.go
```diff
@@ -18,10 +18,17 @@ func TestValidateName_EdgeCases(t *testing.T) {
...
+		{"valid apex", "@", false},
+		{"valid wildcard", "*", false},
+		{"valid subdomain", "sub.domain", false},
 		{"single char label", "a", false},
 		{"two single char labels", "a.b", false},
 		{"exact max length 253", name253, false},
+		{"empty", "", true},
 		{"over max length 254", name254, true},
+		{"too long", strings.Repeat("a", 254), true},
+		{"starts with dot", ".invalid", true},
```
kurok: Redundant with the `name254` case on line 29 — both trip the `len(name) > 253` guard in `ValidateName` before the hostname pattern is ever reached, so they exercise the same branch with the same error path. Drop this one; the descriptively-named `over max length 254` above already covers it.

```suggestion
		{"starts with dot", ".invalid", true},
```

internal/client/records/alias_test.go
```diff
@@ -44,58 +47,12 @@ func TestALIASRecord_ValidateAliasName(t *testing.T) {
...
-			rec.TTL = tt.ttl
-			err := rec.ValidateTTL()
-			if (err != nil) != tt.wantErr {
-				t.Errorf("ValidateTTL(%d) error = %v, wantErr = %v", tt.ttl, err, tt.wantErr)
-			}
-		})
-	}
-}
+// Shared hostname (Name) and TTL edge cases live in common_test.go.
+// Per-type ValidateName / ValidateTTL wiring is covered by TestALIASRecord_Validate_ValidRecord.
```
kurok: This overstates coverage. `TestALIASRecord_Validate_ValidRecord` only checks that a valid record yields zero errors — it keeps passing even if `r.ValidateName` or `r.ValidateTTL` were silently dropped from the `validators` slice in `alias.go:40-44`. The deleted per-type tests didn't prove wiring either (they invoked the method directly), so nothing is actually lost in the refactor — just don't claim the happy-path test asserts wiring.

Same wording appears in `a_test.go:53`, `aaaa_test.go:53`, `caa_test.go:122`, `cname_test.go:58`, `https_test.go:203`, `mx_test.go:97`, `srv_test.go:211` and `srv_test.go:242`. Rephrase across all of them, or add a minimal wiring assertion (set an invalid `Name`/`TTL` on an otherwise-valid record, call `Validate()`, expect the corresponding error).

```suggestion
// Shared hostname (Name) and TTL edge cases live in common_test.go.
// Per-type ValidateName / ValidateTTL are thin forwarders over the shared
// helpers; no per-type wiring assertion beyond the Validate() happy path.
```

internal/provider/domain_info_data_source_acc_test.go
```diff
@@ -86,23 +85,3 @@ func domainInfoSuspensionChecks() resource.TestCheckFunc {
 	return expectListCount(domainInfoDataSourceName, "suspensions.#", 0)
 }
 
-func TestAccDomainInfo_wrongDomainInfo(t *testing.T) {
```
kurok, line 89: Deleting `TestAccDomainInfo_wrongDomainInfo` leaves `stringvalidator.LengthBetween(4, 255)` on `domain_info_data_source.go:62` without any test guarding it — accidentally removing (or loosening) the validator in a future refactor would start silently accepting 3-char domains. If the goal is to stop exercising Terraform Plugin Framework internals, prefer a cheap unit test that instantiates the data-source schema and asserts the expected `validator.String` is registered on the `domain` attribute, rather than dropping the coverage entirely.

kurok (commented):
Nice consolidation — removes genuine duplication. Three things to reconsider before merging (inline). Nit: the trailing `// Shared ... live in common_test.go` signpost is repeated verbatim in 7 files; a single note in `common_test.go` would be cheaper to maintain.

kurok:
Summary of findings (all should-fix, three inline suggestions + one nit in the review body):                                                       
  
  1. common_test.go:30-31 — duplicate "too long" case is fully subsumed by "over max length 254" (both trip the length guard first). Suggestion      
  deletes the redundant line.                                                                                                                      
  2. alias_test.go:57-58 (and 7 other test files) — the trailing comment claims TestXxxRecord_Validate_ValidRecord covers ValidateName/ValidateTTL   
  wiring; it doesn't — that test passes even if the methods are dropped from the validators slice. Suggestion rephrases; noted the other files to    
  fix.
  3. domain_info_data_source_acc_test.go (deleted TestAccDomainInfo_wrongDomainInfo) — leaves stringvalidator.LengthBetween(4, 255) on               
  domain_info_data_source.go:62 without a guard. Recommended a cheap schema-instantiation unit test rather than dropping coverage.                   
  4. Nit (in review body) — the 2-line "Shared ... live in common_test.go" signpost is duplicated across 7 files; one note in common_test.go would be
   cheaper to maintain.

internal/provider/domain_info_data_source_acc_test.go
dm1tr1yvovk: it just tests that terraform schema validation works, nothing provider specific, that is because I've decided to delete it.

PR #70, opened 2026-04-23
feat: add validation for ns records

dm1tr1yvovk:


internal/client/records/ns_test.go
```diff
@@ -0,0 +1,58 @@
...
+			}
+			if tt.wantErrContains != "" && (err == nil || !strings.Contains(err.Error(), tt.wantErrContains)) {
+				t.Errorf("ValidateNameserver(%q) error = %v, want substring %q", tt.nameserver, err, tt.wantErrContains)
+			}
+		})
+	}
+}
+
+// Shared hostname (Name) and TTL edge cases live in common_test.go.
+// Per-type ValidateName / ValidateTTL wiring is covered by TestNSRecord_Validate_ValidRecord.
```
kurok: Remove — this signpost was deleted from every other record test file in #69 (merged commit `7c0bdb8a`) because the wiring claim overstates coverage: `TestNSRecord_Validate_ValidRecord` only verifies that a valid record yields zero errors, and keeps passing even if `r.ValidateName` / `r.ValidateTTL` are silently dropped from the `validators` slice in `ns.go`. The single accurate note in the merged `common_test.go` already covers all record types.

Also fixes the missing trailing newline (`gofmt -l` flag).

```suggestion
}
```

kurok (commented):
Adds NS record support following the established three-file pattern, plus a rename pass dropping the redundant `_record` from validator filenames. The API-probe acceptance test (`TestAccDNSRecords_nsRecordAPIRejectsApexAndWildcard`) for the `@`/`*` pre-check is a nice touch.

Most of this PR (test refactor + `domain_info_data_source_acc_test.go` deletion) overlaps with work merged in #69 — a rebase will cleanly resolve the refactor half on its own.

Major items inline:
1. **gofmt drift** on `ns.go`, `ns_test.go`, `record_ns_validator.go` — all three are missing trailing newlines. `gofmt -l` flags them; CI will fail. Fix with `gofmt -w internal/`.
2. **Stale trailing comment in `ns_test.go`** — the merged #69 removed this exact comment from every other record test file and consolidated an accurate version into `common_test.go`.

Not flagged but worth a thought: no acceptance test exercises NS at the zone apex (`name = "@"`), which is NS's canonical use case. Presumably deliberate — exercising it would overwrite SPACESHIP_TEST_DOMAIN's real delegation. If so, worth a one-line comment at the top of the acc test file noting the constraint so a future contributor doesn't try to add it and break the test domain.

internal/client/records/ns.go
```diff
@@ -0,0 +1,50 @@
...
+		r.ValidateName,
+		r.ValidateTTL,
+	}
+	for _, v := range validators {
+		if err := v(); err != nil {
+			errs = append(errs, err)
+		}
+	}
+	return errs
+}
```
kurok: Missing trailing newline at EOF — `gofmt -l` flags this file (also `ns_test.go` and `record_ns_validator.go`). CI will fail. Run `gofmt -w internal/client/records/ns.go internal/client/records/ns_test.go internal/provider/records/record_ns_validator.go` to fix all three in one pass.

```suggestion
	return errs
}
```

dm1tr1yvovk:
Fair concern about NS, but the premise doesn't hold — overwriting the test domain's NS isn't a problem on our end, so there's nothing to warn contributors off. The real gap is coverage: zone-apex NS is NS's canonical use case and we currently have zero evidence it works through the provider. I'll add an apex NS lifecycle test (name = "@") instead.

internal/provider/records/record_ns_validator.go
```diff
@@ -0,0 +1,71 @@
...
+		rec.Nameserver = nameserverAttr.ValueString()
+		if err := rec.ValidateNameserver(); err != nil {
+			resp.Diagnostics.AddAttributeError(
+				req.Path.AtName("nameserver"),
+				"Invalid Nameserver Value",
+				err.Error(),
+			)
+		}
+	}
+}
```
kurok: Last of the three files still missing the trailing newline (`ns.go` and `ns_test.go` got fixed). `gofmt -w internal/provider/records/record_ns_validator.go` to close this out.

```suggestion
}
```

kurok (commented):
Thanks for the rebase — refactor half dropped out cleanly. One of the three gofmt-flagged files still needs the trailing newline (inline). On the apex-NS acc test: sounds good, looking forward to it.

PR #71, opened 2026-04-24
feat: add validation for ptr records

dm1tr1yvovk:


kurok (commented):
Overall the PTR validator is well-structured and follows the established patterns in this codebase. One correctness gap worth addressing before merge.

**Finding (medium): wildcard-prefixed pointer targets are not blocked**

`ValidatePointer` rejects bare `"*"` but does not reject `"*.example.com"`. The shared `hostnamePattern` in `ValidateName` explicitly permits the `[_*].` prefix (used for wildcard record *names*). A PTR pointer target must be a fully qualified real hostname — wildcards in the target position are invalid per RFC 1035. The pre-check needs to be extended, and the unit test table needs a matching case.

Inline suggestions below.

internal/client/records/ptr.go
```diff
@@ -0,0 +1,50 @@
...
+	Pointer string
+	Name    string
+	TTL     int
+}
+
+// ValidatePointer checks that the pointer target is a valid domain name.
+// "@" and "*" are rejected: a PTR target must be a real hostname, not the
+// apex placeholder or a wildcard.
+func (r *PTRRecord) ValidatePointer() error {
+	if r.Pointer == "@" || r.Pointer == "*" {
```
kurok: The pre-check only blocks bare `"*"`. The shared `hostnamePattern` used by `ValidateName` permits `[_*]\.` prefixes, so `"*.example.com"` passes `ValidatePointer` today. A PTR pointer target must be a real hostname, not a wildcard. Extend the guard:

```suggestion
	if r.Pointer == "@" || r.Pointer == "*" || strings.HasPrefix(r.Pointer, "*.") {
		return fmt.Errorf("must be a valid domain name, got %q", r.Pointer)
	}
```

(Add `"strings"` to the import if not already present.)

internal/client/records/ptr_test.go
```diff
@@ -0,0 +1,107 @@
...
+	}{
+		{"valid hostname", "host.example.com", false, ""},
+		{"valid subdomain", "sub.domain", false, ""},
+		{"valid single label", "myhost", false, ""},
+		{"apex rejected", "@", true, "domain name"},
+		{"wildcard rejected", "*", true, "domain name"},
+		{"empty", "", true, ""},
+		{"too long", strings.Repeat("a", 254), true, ""},
+		{"starts with dot", ".invalid", true, ""},
+	}
```
kurok, line 41: Add a test case for wildcard-prefixed pointer targets alongside the existing `"*"` case:

```suggestion
		{"apex rejected", "@", true, "domain name"},
		{"wildcard rejected", "*", true, "domain name"},
		{"wildcard prefix rejected", "*.example.com", true, "domain name"},
```

PR #72, opened 2026-04-24
feat: add validation and tests for svcb record

dm1tr1yvovk:


kurok (commented):
## Review summary

Two major findings.

### 1. TLSA validation regression (high)

The PR removes the attribute-level `portStringPattern` and `schemeLabelPattern` validators from `port`, `scheme`, and `protocol` fields. This is fine for HTTPS and SRV — `HTTPSValidator` and `SRVValidator` cover those types. But TLSA records use all three fields (`port`, `scheme`, `protocol`) and there is no `TLSAValidator`. Those fields are now completely unvalidated for TLSA records at plan time, whereas they were validated before this PR.

See inline comment on the `protocol` field.

### 2. Spurious scheme error when port fails validation (medium)

In `record_svcb_validator.go`, `rec.Port` is populated with the raw (potentially invalid) value before port validation runs. When `ValidatePort` returns an error, `rec.Port` still holds the bad value. The subsequent `ValidateScheme` call sees a non-empty `rec.Port` and emits a misleading secondary error: _"scheme is required when port is set"_ — even though the port itself was already rejected. A user setting e.g. `port = "_0"` with no scheme will get two confusing errors instead of one.

See inline suggestion on the port block. (Note: `HTTPSValidator` contains the same pattern; fixing it there separately is out of scope for this PR.)

internal/provider/records/record_svcb_validator.go
```diff
@@ -0,0 +1,151 @@
...
+	} else if !portAttr.IsNull() && !portAttr.IsUnknown() {
+		rec.Port = portAttr.ValueString()
+		if err := rec.ValidatePort(); err != nil {
+			resp.Diagnostics.AddAttributeError(
+				req.Path.AtName("port"),
+				"Invalid Port Value",
+				err.Error(),
+			)
+		}
+	}
```
kurok, line 131: Reset `rec.Port` after a failed validation so that `ValidateScheme` does not see a bad port value and emit a spurious "scheme is required when port is set" error alongside the port error.

```suggestion
	} else if !portAttr.IsNull() && !portAttr.IsUnknown() {
		rec.Port = portAttr.ValueString()
		if err := rec.ValidatePort(); err != nil {
			rec.Port = ""
			resp.Diagnostics.AddAttributeError(
				req.Path.AtName("port"),
				"Invalid Port Value",
				err.Error(),
			)
		}
	}
```

internal/provider/dns_records_resource.go
```diff
@@ -321,9 +311,6 @@ func (r *dnsRecordsResource) Schema(_ context.Context, _ resource.SchemaRequest,
 						"protocol": schema.StringAttribute{
```
kurok, line 311: `SRVValidator` covers `protocol` for SRV records, but TLSA records also use `port`, `scheme`, and `protocol` (see lines 786–800 of this file) and there is no `TLSAValidator`. Removing the attribute-level `schemeLabelPattern` validator from this field drops the only validation TLSA had for `protocol`. Either add a `TLSAValidator` that validates these three fields, or restore the attribute-level validators for `port`, `scheme`, and `protocol` (scoped to TLSA via an `AlsoRequires`/type-guard approach similar to the other validators).

dm1tr1yvovk: thank for pointing, it is intentionally, I will add tlsa validator in the next one or two PRs.

kurok (approved):

PR #73, opened 2026-04-27
feat: implement tlsa record creation and validation

dm1tr1yvovk:


kurok:
## Review summary

The PR adds TLSA validation following the established three-file pattern (client struct + object validator + acc tests). Implementation is consistent with the SRV/SVCB validators. `gofmt`, `go vet`, and `go test ./internal/client/records/ -run TestTLSA` are clean.

### Major

**1. Two `tlsaAssociationPattern` vars with different rules** — `internal/provider/dns_records_resource.go:36` (pre-existing, schema-level) and `internal/client/records/tlsa.go:13` (new, client-level) share the name but disagree on case:

```go
// schema (pre-existing) — accepts mixed case
tlsaAssociationPattern = regexp.MustCompile(`^[0-9a-fA-F]{2}(\s?[0-9a-fA-F]{2})*$`)

// client validator (new) — rejects uppercase, also pulls in regexp2 for one lookahead
tlsaAssociationPattern = regexp2.MustCompile(`^(?!\s)(\s?[0-9a-f]{2})+$`, regexp2.None)
```

The `uppercaseassoc` acc test confirms the layered behavior (schema passes → validator rejects). Pick one source of truth. If the API rejects uppercase (which the test implies), tighten the schema regex to lowercase and drop `regexp2` from the new file — the schema regex shape already has the leading-pair anchor without needing a lookahead:

```go
// internal/client/records/tlsa.go (replace lines 8 and 13)
//   - drop "github.com/dlclark/regexp2"
tlsaAssociationPattern = regexp.MustCompile(`^[0-9a-f]{2}(\s?[0-9a-f]{2})*$`)

// in ValidateAssociationData — drop the err return from MatchString
if !tlsaAssociationPattern.MatchString(r.AssociationData) {
    return fmt.Errorf("must be lowercase hex byte pairs optionally separated by single spaces, got %q", r.AssociationData)
}
```

```go
// internal/provider/dns_records_resource.go:36 — align to lowercase too
tlsaAssociationPattern = regexp.MustCompile(`^[0-9a-f]{2}(\s?[0-9a-f]{2})*$`)
```

If the API actually accepts uppercase, do the inverse: drop the lowercase restriction in the client regex (and the test). Either way, the two regexes must agree.

### Should fix

- **`\s` is broader than the spec implies.** `tlsa.go:13` matches tabs/newlines between byte pairs. If the API only allows ASCII spaces, use `[ ]?` (or a literal `' '`) instead of `\s?` to keep the validator faithful to the wire format. The same applies to the schema regex.
- **Length error message is in characters, not byte-pairs.** `ValidateAssociationData` says "must be between 64 and 65535 characters" but counts whitespace too, so a user with 31 pairs + 31 spaces (93 chars) gets a length-pass but regex-fail with no hint that the *byte-pair* count was the real issue. Consider rewording: `"must contain at least 32 hex byte pairs (got %d characters including separators)"`.
- **Test boundary off-by-one.** `TestTLSARecord_ValidateAssociationData` covers 62 chars (too short) and 65534 chars (valid max) but skips 63 (still too short) and 65535 (valid upper bound). Tighten the boundaries — these are exactly the inputs that catch `<=` vs `<` slips.

### Nits

- Same identifier `tlsaAssociationPattern` in two packages with different regexes is a grep hazard. Rename the new one to e.g. `tlsaAssociationDataPattern` even if the regexes are aligned.
- `ValidatePort`'s "got %q" path covers empty string, no-underscore, non-digit suffix, and out-of-range with the same message. Splitting "must start with `_`" from "must be in range 1-65535" gives the user a sharper hint, but it's not load-bearing.
- `validTLSARecord()` in `tlsa_test.go` is good; consider also adding a happy-path table case where `AssociationData` is exactly 65535 chars (current upper bound) to lock the inclusive boundary.

Nothing here is blocking — the core logic mirrors the API constraints and the acc-test matrix is solid (lifecycle, multi-host, boundary, import-pre-existing). Just want the two regexes reconciled before merge.

kurok (changes requested):
Requesting changes — see review summary in https://github.com/namecheap/terraform-provider-spaceship/pull/73#issuecomment-4327647741. The major item (divergent `tlsaAssociationPattern` regexes between schema and client validator) needs to be reconciled before merge.

dm1tr1yvovk:
Thanks for the thorough review @kurok — most of this is in. One twist: we ran an API probe that flipped our direction on the Major item.

## Probe findings (TLSA `associationData` against the live API)

Wrote a discovery test bypassing both validators and hitting the API directly with a matrix of inputs. Data:

| Input | API verdict | Storage |
|---|---|---|
| Lowercase 64-char hex | ✅ accepted | verbatim |
| **Uppercase 64-char hex** | ✅ **accepted** | **verbatim** |
| Mixed case | ✅ accepted | verbatim |
| Single space between byte pairs | ✅ accepted | verbatim |
| Single tab between pairs | ✅ accepted | verbatim |
| Single LF (`\n`) between pairs | ✅ accepted | verbatim |
| CRLF (`\r\n`) between pairs | ❌ 422 "Not a valid Hexadecimal" | — |
| Double space between pairs | ❌ 422 | — |
| Leading whitespace (space or tab) | ❌ 422 | — |
| Trailing whitespace | ❌ 422 | — |

Two surprises: (1) the API accepts all hex casing — the spec doc's `[0-9a-f]` is more restrictive than reality; (2) the API stores the value **verbatim** with no normalization.

Per our project rule "mirror the API, don't invent stricter rules", this means the *correct* reconciliation is to align both regexes to **case-insensitive**, not lowercase. Pattern is now `^[0-9a-fA-F]{2}(\s?[0-9a-fA-F]{2})*$` in both places.

## Per-comment response

| # | Comment | Status |
|---|---|---|
| Major | Reconcile two `tlsaAssociationPattern` vars | ✅ Done — both case-insensitive (probe-justified). Dropped `regexp2` from `tlsa.go` |
| Should-fix #1 | Tighten `\s?` to `[ ]?` | Skipped — probe shows API accepts single tab and LF; tightening would invent a stricter rule than the API enforces |
| Should-fix #2 | Length error message rewording | Skipped as marginal |
| Should-fix #3 | Test boundaries 63 / 65535 | ✅ Done |
| Nit #1 | Rename to break grep hazard | ✅ Done — client-side is now `tlsaAssociationDataPattern`; the two are no longer same-name |
| Nit #2 | Split `ValidatePort` error messages | Skipped — matches SVCB style; you marked it "not load-bearing" |
| Nit #3 | 65535 happy path | ✅ Done (subsumed by #4) |

Also kept the diff-signature normalization at `dns_records_resource.go:1033` (`ToLower` + space-strip) — without it, two semantically identical inputs with different casing would be treated as distinct records during upsert/delete since the API stores verbatim.

kurok (approved):
Major resolved, nits addressed. The  decision is justified by the API probe data. LGTM.

kurok (approved):
Major resolved, nits addressed. The \s? decision is justified by the API probe data. LGTM.

PR #74, opened 2026-04-28
feat: add TXT record validation checks

dm1tr1yvovk:


kurok (commented):
## Summary

**TXT record validation** — adds `TXTRecord` struct with `ValidateValue` (length 1–65535 bytes, rejects whitespace-only), wires it into a Terraform object-level validator, and ships a full acceptance-test suite (10 scenarios: basic round-trip, import, whitespace-only rejection, leading/trailing whitespace passthrough, empty, missing, update, case-sensitivity, multi-value same host, long value, special chars).

**Validator consolidation** — removes field-level attribute validators for CAA `flag`, SRV `priority`/`weight`/`port_number`, and TLSA `usage`/`selector`/`matching`/`association_data`, along with the `tlsaAssociationPattern` regex. Verified these are all covered by the existing object-level validators (`CAAValidator`, `SRVValidator`, `TLSAValidator`) registered in the schema.

## Findings

**Nit — `record_txt_validator.go:22`: Description string is incomplete**
> `"validates that TXT records contain a value within the allowed length"`

The validator also rejects whitespace-only values, which isn't captured here. This is the string shown to users in Terraform provider documentation.

No must-fix or should-fix items found. Rest of the code is clean — good test coverage, idiomatic use of the framework validator pattern, well-placed API-quirk comments in `txt.go`.

kurok (approved):
LGTM. One nit on the Description string (whitespace-only rejection isn't mentioned), but nothing blocking.

PR #75, opened 2026-04-30
feat: add single domain resource record

dm1tr1yvovk:


kurok:
## Code Review Summary

**Must Fix**
- `internal/client/dns.go:151` — parameter name typo: `recoredType` → `recordType`. Public function, shows in docs/IDE.
- `internal/client/dns.go:158` — capitalized error string: `"Not found"` must be lowercase → `"record not found"` (Go error convention).
- `internal/provider/dns_record_resource.go:165` — `Read()` is empty. Resource is registered in production via `provider.go`. Without Read, Terraform cannot refresh state — every plan will show a diff and `terraform import` is broken. Must fetch the record and call `resp.State.Set`, or remove the resource from the provider registration until it's ready.
- `internal/provider/dns_record_resource.go:169` — `Update()` is empty. Any configuration change will silently succeed with no effect, leaving state and reality out of sync.

**Should Fix**
- `internal/provider/dns_record_resource.go:96+100` — `req.Plan.Get(ctx, &plan)` is called twice with no mutation in between; the second call is dead code. Remove the duplicate.
- `internal/provider/dns_record_resource.go:128` — `plan.ID = types.StringValue(domain)` sets a non-unique ID. Multiple A records on the same domain will collide in state. Use the composite key `domain + "/" + type + "/" + name + "/" + value` (see `recordKey()` in `dns_records_resource.go`).
- `internal/provider/dns_records_resource_acc_test.go` — three identical helpers: `intPointerClient`, `intPointer`, `intPtr`. TODO comment already says to leave only one. Consolidate before this merges.

**Nits**
- `internal/client/dns.go` — commented-out `DeleteDNSRecord` block (old PUT-based implementation) should be deleted, not committed.
- `internal/provider/dns_record_resource.go` — exploratory comments (`// this line is super important`, `// how it even works?`, `// how to create any other record field?`) are development notes; remove before merging.
- `internal/provider/record_attributes.go:15` — doc comment typo: `retrusn` → `returns`.

internal/client/dns.go
```diff
@@ -149,6 +149,20 @@ func (c *Client) GetDNSRecords(ctx context.Context, domain string) ([]DNSRecord,
 	return filterCustomDNSRecords(result), nil
```
kurok, line 108: Typo in parameter name: `recoredType` → `recordType`. This is a public function so the parameter name appears in godoc and IDE completions.

internal/client/dns.go
```diff
@@ -149,6 +149,20 @@ func (c *Client) GetDNSRecords(ctx context.Context, domain string) ([]DNSRecord,
...
 }
 
+func (c *Client) GetSpecificDNSRecord(ctx context.Context, domain string, recoredType string, recordName string) (DNSRecord, error) {
+	records, err := c.GetDNSRecords(ctx, domain)
+	if err != nil {
+		return DNSRecord{}, err
+	}
+	for _, record := range records {
+		if record.Type == recoredType && record.Name == recordName {
```
kurok: Capitalized error string violates Go convention. Use lowercase, no trailing punctuation: `fmt.Errorf("record not found")`. Consider a package-level sentinel: `var ErrRecordNotFound = errors.New("record not found")` so callers can use `errors.Is`.

internal/provider/dns_record_resource.go
```diff
@@ -0,0 +1,184 @@
...
+	// how to create any other record field?
+
+	address := state.Address.ValueString()
+	record := client.DNSRecord{
+		Type:    recordType,
+		Name:    name,
+		TTL:     3600,
+		Address: address,
+	}
+
```
kurok, line 195: `Read` is empty but the resource is registered in `provider.go`. This breaks `terraform refresh`, `terraform plan` (perpetual diff), and `terraform import`. Minimum implementation: call `r.client.GetSpecificDNSRecord`, set state, and call `resp.State.RemoveResource` if the record is gone (404-style). If not ready to ship, remove `NewDNSRecordResource` from the provider registration.

internal/provider/dns_record_resource.go
```diff
@@ -0,0 +1,184 @@
...
+	if err := r.client.DeleteDNSRecord(ctx, domain, record); err != nil {
+		resp.Diagnostics.AddError("Spaceship API error", fmt.Sprintf("Failed to delete DNS record: %s", err))
+		return
+	}
```
kurok, line 199: `Update` is empty. Any change to a managed record's attributes will appear to succeed but will not be applied. DNS records in Spaceship are upserted by PUT, so Update can delegate to `r.client.CreateDNSRecord` (which uses PUT semantics) after reading new plan values — same as Create, plus refreshing state.

kurok (changes requested):
4 must-fix items: empty Read/Update on a production-registered resource, capitalized error string, and a typo in a public function parameter. Details in the summary comment and inline annotations.

dm1tr1yvovk:
@kurok thank you for review, the feature is not complete at all, I've just decided to publish current state of work before my vacation and made it as draft. feature is not ready for review.

kurok:
### Code review

Found 3 issues:

1. `requireString` and `requireInt` emit a \"missing field\" error for `IsUnknown()` values, causing phantom plan-time errors when record attributes reference computed outputs from other resources. Unknown values should be allowed through without error; only `IsNull()` should fail.

https://github.com/namecheap/terraform-provider-spaceship/blob/33740ba543abad8b17bf3b416fd58099ecc6882c/internal/provider/dns_record_common.go#L224-L237

2. `parseRecordID` validates `parts[0]`–`parts[2]` but not `parts[3]` (the data signature). An import ID with a trailing slash (e.g. `example.com/TXT/@/`) returns `ok=true` with an empty signature, causing `FindDNSRecord` to match nothing and silently remove the resource from state on the next Read.

https://github.com/namecheap/terraform-provider-spaceship/blob/33740ba543abad8b17bf3b416fd58099ecc6882c/internal/provider/dns_record_resource.go#L300-L307

3. `RecordValueSignature` for `SRV` only includes `service`, `protocol`, `priority`, and `weight` — `port` and `target` are excluded. Two SRV records differing only in `target` or `port` share the same signature and therefore the same composite ID. Since both fields carry `RequiresReplace`, a change triggers a replace, but the new resource gets the same ID as the destroyed one; on re-import the wrong record could be matched.

https://github.com/namecheap/terraform-provider-spaceship/blob/33740ba543abad8b17bf3b416fd58099ecc6882c/internal/client/dns_match.go#L48-L51

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

kurok:
## Code Review: feat: add single domain resource record

### Summary

Adds `spaceship_dns_record` (singular) alongside `spaceship_dns_records` (plural). Core refactor: extracts `dnsRecordModel`, `hydrateRecordModel`, `modelToDNSRecord`, `recordAttributes()`, and the signature/key logic into shared files, moves `PortValue` to its own file and match functions to the `client` package. The new singular resource covers all DNS types, handles TTL-only in-place updates, replace on data changes, import, and ConfigValidator adaptation.

---

### Findings

#### Must Fix

- **`internal/provider/dns_record_common.go:162` — TODO comment left in production code**
  ```
  // TODO
  // why it is stringnull?
  ```
  The code is correct; remove the comment.

- **`internal/provider/dns_record_common.go:185,189` — Two more inline TODOs**
  ```
  // TODO
  // WHY it statially defined here and not in model for all records?
  // TTL FIeld is not required
  // what is default on the site
  ```
  and
  ```
  // TODO
  // why to convert here again?
  ```
  Remove before merging.

- **`internal/provider/dns_records_resource.go:33` — TODO left in production file**
  ```
  // TODO move this validator in separate file?
  ```
  Remove or act on it.

- **`internal/provider/dns_records_resource.go:469` — TODO block left in production file**
  ```
  // TODO
  // move to some other file
  ```
  Remove.

- **`internal/client/dns.go:196` — `DeleteDNSRecord` is a pointless wrapper with redundant `err != nil` pattern**

  Current:
  ```go
  func (c *Client) DeleteDNSRecord(ctx context.Context, domain string, record DNSRecord) error {
      records := []DNSRecord{record}
      err := c.DeleteDNSRecords(ctx, domain, records)
      if err != nil {
          return err
      }
      return nil
  }
  ```
  Should be:
  ```go
  func (c *Client) DeleteDNSRecord(ctx context.Context, domain string, record DNSRecord) error {
      return c.DeleteDNSRecords(ctx, domain, []DNSRecord{record})
  }
  ```

- **`internal/client/dns.go:167-178` — `CreateDNSRecord` has the same `if err != nil { return err }; return nil` pattern**

  The last block should be `return c.doJSON(...)` directly. Also the `records` local variable is used only once — inline it into the payload struct literal.

- **`internal/provider/record_attributes.go:14` — malformed doc comment on `recordAttributes`**
  ```
  // record attributes retrusn the schema attributes that describe a single DNS record.
  ```
  "retrusn" is a typo, and Go convention requires comments to start with the name of the thing being described: `// recordAttributes returns …`

#### Should Fix

- **`internal/client/dns_match_test.go:7` — `// TODO reuse existing` comment left in**
  The `intPtr` helper being in a different package (`client` vs `provider`) means a local copy is fine, but the TODO must be removed.

- **`internal/provider/dns_record_resource.go:45-52` — `domain` has `UseStateForUnknown` plan modifier**
  `domain` is `Required`, so it is always known from config — this modifier is a no-op. Remove it.

- **`internal/provider/dns_record_resource.go:263` — stale reference in `recordID` doc comment**
  The comment mentions `recordKey()` (the deleted local function). Update to reference `client.RecordKey`.

- **`internal/provider/dns_record_common.go:59-91` — field comments duplicate schema descriptions**
  The inline struct-field comments (`// IPv4 or IPv6 address for A and AAAA records`) duplicate the `MarkdownDescription` fields in `recordAttributes()`. They will drift. Remove them or keep only the non-obvious ones.

- **`internal/client/dns_match.go:21` — `write` closure is variadic but always called with one argument**
  Every call site passes exactly one `string`. Either make it non-variadic (`write(part string)`) to make the contract explicit, or document why variadic was chosen.

#### Nits

- `internal/provider/record_attributes.go:152` — `recordTypeObjectValidators()` comment starts lowercase; fix to `// recordTypeObjectValidators returns …`
- `internal/client/dns.go:133` — trailing blank line before `}` in `FindDNSRecord`; remove it.
- `internal/provider/dns_record_resource.go` — `singularRecordValidator` (50+ lines) would be cleaner in its own file, e.g. `dns_record_validator.go`.

---

### Automated Checks

- ✅ `gofmt -l` — clean
- ✅ `go vet ./...` — clean

kurok (commented):
Follow-up pass — prior round's items (empty Read/Update, ErrRecordNotFound sentinel, recoredType typo) all look addressed. Five new findings below.

internal/client/dns_match_test.go
```diff
@@ -0,0 +1,162 @@
+package client
+
+import "testing"
+
+// TODO reuse existing
```
kurok: Unresolved TODO: `intP` already exists in `dns_port_test.go` (same package). Either rename `intP` → `intPtr` there and drop this duplicate, or remove the comment. Shipping a `// TODO reuse existing` is incomplete cleanup.

internal/client/dns_match_test.go
```diff
@@ -0,0 +1,162 @@
...
+		{"unknown", DNSRecord{Type: "UNKNOWN", Name: "@", Address: "fallback"}},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			sig := RecordValueSignature(tc.record)
+			if sig == "" && tc.name != "unknown" {
+				t.Error("expected non-empty signature")
+			}
+			if sig != RecordValueSignature(tc.record) {
```
kurok: Vacuous assertion — `RecordValueSignature` is a pure function, so calling it twice with the same input is always equal and this branch can never fire. Remove it, or if you want a genuine determinism smoke-test, make two independent calls and compare them (though for a pure function that still proves very little).

internal/provider/record_attributes.go
```diff
@@ -0,0 +1,166 @@
...
+	"terraform-provider-spaceship/internal/provider/records"
+
+	"github.com/hashicorp/terraform-plugin-framework-validators/int64validator"
+	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/int64default"
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+)
+
+// record attributes retunn the schema attributes that describe a single DNS record.
```
kurok: Two issues: (1) typo "retunn" → "return"; (2) Go doc-comment convention requires the comment to start with the function name. Suggested replacement:\n\n```go\n// recordAttributes returns the schema attributes that describe a single DNS record.\n// Shared by the list resource and the singular dns_record resource;\n// the caller merges in id and domain at the resource root.\n```

internal/client/dns.go
```diff
@@ -185,6 +167,32 @@ func (c *Client) UpsertDNSRecords(ctx context.Context, domain string, force bool
...
+		Items []DNSRecord `json:"items"`
+	}{
+		Items: []DNSRecord{record},
+	}
+
+	_, err := c.doJSON(ctx, http.MethodPut, endpoint, payload, nil)
+	return err
+}
+
+func (c *Client) DeleteDNSRecord(ctx context.Context, domain string, record DNSRecord) error {
```
kurok, line 194: `DeleteDNSRecord` is exported but has no doc comment. The sibling `CreateDNSRecord` has a detailed one. Add at minimum:\n\n```go\n// DeleteDNSRecord removes a single custom DNS record from the domain.\n```

internal/provider/dns_record_resource.go
```diff
@@ -0,0 +1,323 @@
...
+}
+
+func (r *dnsRecordResource) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
+	if r.client == nil {
+		resp.Diagnostics.AddError("Unconfigured provider", "The Spaceship provider was not configured. Please ensure the provider block is present")
+		return
+	}
+
+	var plan dnsRecordResourceModel
+	resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
```
kurok, line 118: Missing guard after `Plan.Get`. `Delete`, `Read`, and `Update` all check `resp.Diagnostics.HasError()` immediately after their state/plan read; `Create` is the odd one out. If `Plan.Get` fails, `domain` is empty and `modelToDNSRecord` emits further cascading errors before the real guard downstream. Suggested fix:\n\n```go\nresp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)\nif resp.Diagnostics.HasError() {\n\treturn\n}\n```

internal/provider/dns_record_common.go
```diff
@@ -0,0 +1,382 @@
...
+import (
+	"fmt"
+	"strings"
+	"terraform-provider-spaceship/internal/client"
+
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
```
kurok, line 13: **Must Fix — import grouping: internal package mixed with stdlib group**

The `terraform-provider-spaceship/internal/client` import is in the same block as stdlib packages (`"fmt"`, `"strings"`). `goimports` requires a blank line between stdlib and non-stdlib imports.

```suggestion
import (
	"fmt"
	"strings"

	"terraform-provider-spaceship/internal/client"

	"github.com/hashicorp/terraform-plugin-framework/attr"
	"github.com/hashicorp/terraform-plugin-framework/diag"
	"github.com/hashicorp/terraform-plugin-framework/path"
	"github.com/hashicorp/terraform-plugin-framework/types"
)
```

internal/provider/dns_record_resource.go
```diff
@@ -0,0 +1,326 @@
...
+	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/int64planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
```
kurok, line 21: **Must Fix — import grouping: internal package mixed with stdlib group**

`"terraform-provider-spaceship/internal/client"` sits in the same block as `"context"`, `"errors"`, `"fmt"`, `"maps"`, `"strings"`. Needs its own group separated from stdlib.

```suggestion
import (
	"context"
	"errors"
	"fmt"
	"maps"
	"strings"

	"terraform-provider-spaceship/internal/client"

	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
	"github.com/hashicorp/terraform-plugin-framework/path"
	"github.com/hashicorp/terraform-plugin-framework/resource"
	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
	"github.com/hashicorp/terraform-plugin-framework/resource/schema/int64planmodifier"
	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
	"github.com/hashicorp/terraform-plugin-framework/types"
)
```

internal/client/dns_match_test.go
```diff
@@ -0,0 +1,157 @@
...
+				t.Error("signature should be deterministic")
```
kurok: **Should Fix — vacuous assertion**

`RecordValueSignature` is a pure function: calling it twice with the same argument always returns the same value, so this `if` can never be true. The assertion is dead code and will never catch a regression.

To actually test determinism, call it on a second independent copy of the record, or simply drop the assertion — the non-empty check on the line above is the useful guard.

```suggestion
			sig2 := RecordValueSignature(tc.record)
			if sig != sig2 {
				t.Errorf("RecordValueSignature is not deterministic for %s: first=%q second=%q", tc.name, sig, sig2)
			}
```

internal/provider/record_attributes.go
```diff
@@ -0,0 +1,166 @@
...
+	"github.com/hashicorp/terraform-plugin-framework-validators/int64validator"
+	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/int64default"
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+)
+
+// record attributes return the schema attributes that describe a single DNS record.
+// shared by the list resource and the single-record resource.
+// merged with id + domain at the resource root
```
kurok: **Should Fix — doc comment convention**

Go doc comments must start with the name of the symbol being described, and each sentence must end with a period. The current comment starts with `"record attributes"` (lowercase, space-separated) instead of `"recordAttributes"`, and lines 2–3 have no trailing period.

```suggestion
// recordAttributes returns the schema attributes that describe a single DNS record.
// Shared by the list resource and the singular resource; merged with id and domain at the resource root.
```

kurok (approved):
All review comments addressed. LGTM.

PR #78, opened 2026-06-05
feat: introduce personal nameserver resource

dm1tr1yvovk:


internal/provider/personal_nameserver_resource.go
```diff
@@ -0,0 +1,325 @@
...
+	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
```
kurok, line 22: Should Fix: import groups are ordered stdlib → internal → external. Standard `goimports -local terraform-provider-spaceship` puts external before internal, and the other two new files in this PR (`ip_string_validator.go`, `ip_string_validator_test.go`) already follow that convention. Swap the two middle blocks:

```suggestion
import (
	"context"
	"errors"
	"fmt"
	"strings"

	"github.com/hashicorp/terraform-plugin-framework-validators/setvalidator"
	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
	"github.com/hashicorp/terraform-plugin-framework/diag"
	"github.com/hashicorp/terraform-plugin-framework/path"
	"github.com/hashicorp/terraform-plugin-framework/resource"
	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
	"github.com/hashicorp/terraform-plugin-framework/types"

	"terraform-provider-spaceship/internal/client"
	"terraform-provider-spaceship/internal/provider/records"
)
```

internal/provider/records/ip_string_validator_test.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+// A malformed address must surface a diagnostic so terraform plan catches it
+// before apply.
+func TestIPAddressValidator_RejectsMalformed(t *testing.T) {
+	resp := &validator.StringResponse{}
+	IPAddressValidator().ValidateString(context.Background(),
+		validator.StringRequest{ConfigValue: types.StringValue("not-an-ip")}, resp)
```
kurok: Nit: `context.Background()` is used here while `personal_nameservers_test.go` (same PR) consistently uses `t.Context()`. Prefer `t.Context()` — it cancels automatically when the test finishes and keeps the style uniform. Same change applies to lines 31 and 45.

```suggestion
	IPAddressValidator().ValidateString(t.Context(),
		validator.StringRequest{ConfigValue: types.StringValue("not-an-ip")}, resp)
```

internal/client/personal_nameservers_test.go
```diff
@@ -0,0 +1,176 @@
...
+		ns      PersonalNameserver
+		wantErr bool
+	}{
+		{"valid", PersonalNameserver{Host: "ns1", IPs: []string{"1.2.3.4", "2001:db8::1"}}, false},
+		{"empty host", PersonalNameserver{Host: "", IPs: []string{"1.2.3.4"}}, true},
+		{"apex host rejected", PersonalNameserver{Host: "@", IPs: []string{"1.2.3.4"}}, true},
+		{"wildcard host rejected", PersonalNameserver{Host: "*", IPs: []string{"1.2.3.4"}}, true},
+		{"no ips", PersonalNameserver{Host: "ns1", IPs: nil}, true},
+		{"too many ips", PersonalNameserver{Host: "ns1", IPs: validIPs(17)}, true},
+		{"invalid ip", PersonalNameserver{Host: "ns1", IPs: []string{"not-an-ip"}}, true},
```
kurok, line 163: Nit: the table exercises the lower bound (`Host: ""`) but not the upper bound (256 chars). `ValidateHost` checks `len > 255` — add a case for it:

```suggestion
		{"invalid ip", PersonalNameserver{Host: "ns1", IPs: []string{"not-an-ip"}}, true},
		{"host too long", PersonalNameserver{Host: strings.Repeat("a", 256), IPs: []string{"1.2.3.4"}}, true},
```

internal/provider/personal_nameserver_resource.go
```diff
@@ -0,0 +1,325 @@
...
+	// point at the offending input. Validate them individually rather than via
+	// the aggregate ns.Validate(), whose flat []error drops the field origin.
+	if err := ns.ValidateHost(); err != nil {
+		diags.AddAttributeError(path.Root("host"), "Invalid host", err.Error())
+	}
+	if err := ns.ValidateIPs(); err != nil {
+		diags.AddAttributeError(path.Root("ips"), "Invalid IP addresses", err.Error())
+	}
+
+	return ns, diags
```
kurok, line 267: Nit: the `ElementsAs` error path (line 246) returns `client.PersonalNameserver{}` on failure, but the validation error path here returns the partially-constructed `ns`. Callers guard with `HasError()` so it's harmless, but returning a zero value consistently removes the asymmetry and signals clearly that the return value must not be used on error:

```suggestion

	if diags.HasError() {
		return client.PersonalNameserver{}, diags
	}
	return ns, diags
```

kurok (changes requested):

internal/provider/personal_nameserver_resource.go
dm1tr1yvovk: The previous stdlib → internal → external order was intentional — it came from your own feedback on #75 (the *Must Fix — import grouping* comments on `dns_record_common.go` and `dns_record_resource.go`), and I added the `goimports.local-prefixes` config in commit ecba8fb thinking it would enforce that layout.

You're right that the Go / Terraform-provider community standard is **stdlib → third-party → local module** (what `goimports -local` actually produces, and what `terraform-provider-aws`/`google`/`azurerm` use). Re-litigating this on every PR isn't useful, so I'm switching to align with the community standard now:

- Removed the `goimports.local-prefixes` block from `.golangci.yml`.
- Reordered all 29 affected files to stdlib → external → local.
- This file now matches that order too.

Will resolve once you've had a look.

kurok: Verified at `972a2c2`: `local-prefixes` removed, 29 files reordered, this file matches, and `golangci-lint fmt --diff` is clean across the tree.

And you're right about the #75 inconsistency — those "Must Fix" suggestions did put the local group in the middle (stdlib → internal → external). That earlier guidance was wrong; this is the correct order. Sorry for the churn.

Root-cause note so it doesn't resurface: neither config actually *enforced* group order. `goimports` (with or without `local-prefixes`) only enforces the stdlib vs non-stdlib split — it leaves existing manual grouping alone and won't flag stdlib → local → external. I confirmed that locally. That's why the order drifted between PRs.

If you want it enforced (no more per-PR judgment calls), swap the formatter to `gci`:

```yaml
formatters:
  enable:
    - gofmt
    - gci
  settings:
    gci:
      sections:
        - standard
        - default
        - prefix(terraform-provider-spaceship)
```

`gci` reorders the groups (verified it auto-fixes a misgrouped block), so CI catches regressions. Optional — fine to merge as-is if you'd rather keep it lightweight. Current state is correct either way; resolving from my side.

kurok (approved):
Approving. Verified at `972a2c2`:

- All three nits are addressed in code — resolving those threads.
- `gofmt -l`, `go vet ./...`, `golangci-lint run`, `go build .`, and `make test` are all clean.
- Lifecycle is sound: rename-in-place (PUT path = current host, body = desired host), the `id` plan modifier correctly handles the rename instead of freezing via `UseStateForUnknown`, both not-found paths in `Read` remove the resource, and `Delete` swallows 404. The 501 list+filter workaround is well documented with a clear `TODO(api-501)`.
- Acc tests cover create / IP update / rename / import / destroy plus out-of-band removal, with `ExpectEmptyPlan` as the convergence check.

One non-blocking thought for later: `hydrate` trusts the PUT response to echo `host` back; the rename step's post-refresh `ExpectEmptyPlan` is what guarantees that against the real API. If it ever flakes, deriving `host` from the plan would drop the dependency. Not needed for merge.

PR #81, opened 2026-06-05
chore: improve makefile and running tests locally

dm1tr1yvovk:


.env.example
```diff
@@ -0,0 +1,25 @@
...
+# `make testacc` loads `.env` automatically, so you never have to export these
+# by hand. Never commit `.env` — it holds live credentials.
+
+# Required: Spaceship API credentials.
+SPACESHIP_API_KEY=
+SPACESHIP_API_SECRET=
+
+# Required: a disposable domain that is SAFE TO MODIFY. Acceptance tests
+# create, update, import, and destroy real DNS records on this domain.
+SPACESHIP_TEST_DOMAIN=example.com
```
kurok, line 16: Nit: `example.com` is a real (ICANN-reserved) domain. A user who copies `.env.example` → `.env` without editing it will get confusing API errors rather than an obvious placeholder that signals the value needs filling in.

```suggestion
SPACESHIP_TEST_DOMAIN=your-disposable-domain.tld
```

kurok (approved):

PR #83, opened 2026-06-09
refactor(records): simplify single-field DNS record validators

kurok:
## Summary

Refactor the seven single-field DNS record validators (A, AAAA, ALIAS, CNAME, NS, PTR, TXT) for readability. No behaviour change.

> **Scope note:** an earlier revision of this PR also renamed the TLSA/HTTPS/SVCB diagnostic titles for consistent word spacing. Per review (dm1tr1yvovk), that half was dropped — the previous titles mirror the Go struct field names, the diagnostic `Path` already carries the snake_case schema name, and renaming is an observable breaking change for anyone matching diagnostic text, with no schema/state change to justify it. Only the refactor remains.

## Change

Each of the seven validators allocated an empty client record struct, set its single field, then validated it inside the trailing branch of an `if/else-if/else` chain. The intermediate struct served no purpose beyond that one call.

They now use guard clauses (early return on the type assertion and null/unknown checks) and inline the one-field struct into the validation call.

Behaviour is unchanged: each validator checks a single field, so the early returns are equivalent to the previous trailing `else`, and each still emits exactly one diagnostic. Multi-field validators (CAA, MX, SRV, TLSA, HTTPS, SVCB) are **intentionally left as-is** — they accumulate diagnostics across several fields, so an early-return rewrite would drop errors.

## Verification

- `go build ./...` — ok
- `go vet ./...` — clean
- `gofmt -l internal/` — clean
- `go test ./internal/provider/records/...` (unit) — pass
- `golangci-lint run ./...` — **0 issues**

Acceptance tests not run (hit the real Spaceship API); no acc-test changes in this revision.

### DIFF
--- a/internal/provider/records/record_a_validator.go
+++ b/internal/provider/records/record_a_validator.go
@@ -43,8 +43,6 @@ func (v *aRecordValidator) ValidateObject(ctx context.Context, req validator.Obj
 		return
 	}
 
-	rec := &clientrecords.ARecord{}
-
 	addressAttr, ok := attrs["address"].(types.String)
 	if !ok {
 		resp.Diagnostics.AddAttributeError(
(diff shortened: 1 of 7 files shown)


dm1tr1yvovk:
Thanks for the writeup — the investigation table is appreciated. After looking over both halves, I'd like to go forward with **just the refactor** and drop the rename.

**Keep:** the seven single-field validator simplifications (A, AAAA, ALIAS, CNAME, NS, PTR, TXT). Guard-clause + inline struct is a genuine readability win, behaviour is preserved (each block sits at the function tail, so `else` ⇄ early-return is equivalent), and your reasoning for *not* applying it to multi-field validators is correct — they accumulate diagnostics across fields and would lose errors under early-return.

**Drop:** the diagnostic-title renames and the matching `errRegex` updates in the acc tests.

Reasoning:
- The previous titles (`TargetName`, `SvcPriority`, `AssociationData`) mirror the Go struct field names. The new titles don't map to the schema any *more* faithfully — the schema names are snake_case (`target_name`, etc.), and the diagnostic `Path` already carries the correct snake_case name via `req.Path.AtName(...)`. So both encodings are equally valid renderings; this is style, not a fix.
- It's an observable breaking change for anyone matching diagnostic text in CI scripts or wrappers, with no schema/state change to justify it. release-please would tag this `fix:` (patch), which understates the externally visible churn.
- The rename also leaves CNAME inconsistent: `record_cname_validator.go:67` still emits `"Invalid CName Value"` (joined PascalCase) and the acc test still asserts that form. If we apply the "split joined PascalCase" rule we should hit CNAME too — but I'd rather not pay the churn at all.

kurok:
Done — dropped the diagnostic-title renames and the matching `errRegex` updates; the TLSA/HTTPS/SVCB titles and acc tests are back to the `master` form (CNAME's `Invalid CName Value` likewise untouched). The seven single-field validator simplifications remain. Force-pushed the rebased branch (now a single `refactor:` commit) and updated the PR description to match the narrowed scope.

`go build` / `go vet` / `gofmt -l` clean, unit tests pass, `golangci-lint run ./...` reports 0 issues.

dm1tr1yvovk (approved):

PR #84, opened 2026-06-09
ci: install latest matching Go patch in setup-go

kurok:
## Summary

Add `check-latest: true` to every `actions/setup-go@v6` step so CI installs the latest Go patch matching `go.mod` instead of reusing whatever patch is pre-cached on the runner image.

By default `setup-go` will use a pre-installed Go on the runner if it satisfies the version spec, which can be several patches behind the toolchain `go.mod` requests. `check-latest: true` forces a check for the newest matching release.

## Changed

- `ci.yml` — both jobs (Check, Acceptance Tests)
- `codeql.yml`
- `release.yml`

Each step keeps `go-version-file: go.mod`; only the new line is added.

## Verification

- YAML parses (`yaml.safe_load`) for all three files
- Diff is 4 added lines, no other changes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,6 +30,7 @@ jobs:
               uses: actions/setup-go@v6
               with:
                   go-version-file: go.mod
+                  check-latest: true
 
             - name: Lint
               uses: golangci/golangci-lint-action@v9
@@ -114,6 +115,7 @@ jobs:
               uses: actions/setup-go@v6
               with:
                   go-version-file: go.mod
+                  check-latest: true
 
             - name: Set up Terraform
               if: ${{ steps.acc_secrets.outputs.has_secrets == 'true' }}
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -49,6 +49,7 @@ jobs:
               uses: actions/setup-go@v6
               with:
                   go-version-file: go.mod
+                  check-latest: true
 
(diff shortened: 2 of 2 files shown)


dm1tr1yvovk:
Thanks for the PR — pushing back a bit before we merge.

`go-version-file: go.mod` already pins the floor: anything the runner ships will be `>= 1.25.9`, which is what the code requires. `check-latest: true` only closes the small gap between the runner image's cached Go and the freshest upstream patch — useful in some contexts, but for a Terraform provider it's a marginal security benefit at the cost of an extra network round-trip per job.

The bigger concern is `release.yml`: with `check-latest` on, two re-runs of the same tag could pick up different Go patches, making release binaries non-deterministic. For a published provider artifact that's a property I'd rather not give up without a clear reason.

If we want strict toolchain pinning for releases, the right lever is a `toolchain go1.25.9` directive in `go.mod` (which makes Go itself fetch the exact toolchain) — not `check-latest`, which pulls the other direction.

Proposal: either drop this PR, or scope it to `ci.yml` + `codeql.yml` only and leave `release.yml` deterministic. WDYT?

kurok:
Agreed on the release determinism point. Reverted `check-latest` from `release.yml`; the change is now scoped to `ci.yml` + `codeql.yml` only. Commit message updated to record why `release.yml` is intentionally left untouched (and that a `toolchain` directive is the right lever if we ever want strict release pinning).

dm1tr1yvovk (approved):

PR #88, opened 2026-06-12
feat: dns record read cache

dm1tr1yvovk:
## Why

Every `spaceship_dns_record` resource re-fetched (and re-paginated) the entire zone on refresh, so N records in one domain cost N full zone reads per `terraform plan`/`apply`. With large configs this is slow and burns API rate limit for no reason.

## What

- **Shared per-domain read cache** (`dnsRecordCache`): `Read`/`Update` of `spaceship_dns_record` fetch through it, collapsing N reads into one zone fetch per domain per Terraform command. `singleflight` collapses the cold-start stampede from Terraform's concurrent refresh (default parallelism 10) into a single fetch.
- **Write invalidation**: every write path of the singular resource (create / TTL update / delete) invalidates its domain, so subsequent reads re-fetch. Fetch errors are never cached.
- **`providerData` bundle**: resources and data sources now receive `{Client, DNSRecords}` via `ProviderData` instead of the raw client, hence the mechanical `Configure` changes across all resources.
- **`client.MatchDNSRecord`** extracted from `FindDNSRecord` so the cache reuses the exact API matching rules (case-insensitive except TXT values) without duplicating them.
- `golang.org/x/sync` promoted to a direct dependency.

## Scope notes

- The plural `spaceship_dns_records` resource intentionally does **not** participate in the cache: mixing it with singular records on one domain is already documented as unsupported, so the cache doesn't defend against that configuration.
- The cache is per-process, i.e. naturally scoped to a single Terraform command — no TTL/expiry needed.

## Testing

Unit tests cover cache hits, write-invalidation (including per-domain scope), singleflight collapse under 20 concurrent readers, error-not-cached recovery, and the `ErrRecordNotFound` contract. `make test`, `make lint`, and `go build` all pass.

### DIFF
--- a/internal/provider/dns_record_cache.go
+++ b/internal/provider/dns_record_cache.go
@@ -16,11 +16,15 @@ import (
 // zone, so N records in one domain cost N full zone reads. The cache collapses
 // that into one read per domain.
 //
-// Correctness rests on write-invalidation: any mutation of a domain's records
-// must call Invalidate(domain) so a later Find re-fetches instead of serving
-// stale data. The cache deliberately lives in the provider layer (not the
-// client) so the client stays a cache-free, reusable API surface — which means
-// the client cannot invalidate on its own, and callers own that responsibility.
+// Correctness rests on write-invalidation: every resource that reads through
+// the cache must call Invalidate(domain) after mutating that domain's records,
+// so a later Find re-fetches instead of serving stale data. The plural
+// spaceship_dns_records resource deliberately does not participate (neither
+// reading nor invalidating): mixing it with the singular resource on one
+// domain is documented as unsupported, so the cache does not defend against
+// that configuration. The cache lives in the provider layer (not the client)
+// so the client stays a cache-free, reusable API surface — which means the
+// client cannot invalidate on its own, and callers own that responsibility.
 type dnsRecordCache struct {
 	client *client.Client
 
@@ -79,8 +83,13 @@ func (c *dnsRecordCache) records(ctx context.Context, domain string) ([]client.D
 	// is held only around the map, never across the network call. A caller that
 	// arrives just after a flight completes (and the key is forgotten) simply
 	// runs one extra fetch — never stale, just an occasional redundant read.
-	result, err, _ := c.sf.Do(domain, func() (any, error) {
-		records, err := c.client.GetDNSRecords(ctx, domain)
+	//
+	// DoChan + select lets each caller observe its own context cancellation,
+	// and context.WithoutCancel detaches the shared fetch from any single
+	// caller's ctx so one caller's cancellation can't fail waiters that still
+	// need the result.
+	ch := c.sf.DoChan(domain, func() (any, error) {
+		records, err := c.client.GetDNSRecords(context.WithoutCancel(ctx), domain)
 		if err != nil {
 			return nil, err
 		}
@@ -89,8 +98,14 @@ func (c *dnsRecordCache) records(ctx context.Context, domain string) ([]client.D
 		c.mu.Unlock()
 		return records, nil
 	})
-	if err != nil {
-		return nil, err
+
+	select {
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	case res := <-ch:
+		if res.Err != nil {
+			return nil, res.Err
+		}
+		return res.Val.([]client.DNSRecord), nil
 	}
-	return result.([]client.DNSRecord), nil
 }
--- a/internal/provider/dns_record_cache.go
+++ b/internal/provider/dns_record_cache.go
@@ -1,0 +1,96 @@
+package provider
+
+import (
+	"context"
+	"sync"
+
+	"golang.org/x/sync/singleflight"
+
+	"terraform-provider-spaceship/internal/client"
+)
+
+// dnsRecordCache memoizes per-domain DNS record fetches for the lifetime of a
+// provider process (i.e. a single Terraform command). The singular
+// spaceship_dns_record resource calls Find once per managed record during a
+// refresh; without this cache each call re-fetches and paginates the whole
+// zone, so N records in one domain cost N full zone reads. The cache collapses
+// that into one read per domain.
+//
+// Correctness rests on write-invalidation: any mutation of a domain's records
+// must call Invalidate(domain) so a later Find re-fetches instead of serving
+// stale data. The cache deliberately lives in the provider layer (not the
+// client) so the client stays a cache-free, reusable API surface — which means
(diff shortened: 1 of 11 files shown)


kurok (commented):
Major items only, with code suggestions. The refactor (extracting `MatchDNSRecord`, bundling `providerData`) is clean and behavior-preserving; `gofmt`/`go vet` are clean and the cache tests pass. Two real issues in the new caching layer are inline below.

internal/provider/dns_records_resource.go
```diff
@@ -100,12 +100,12 @@ func (r *dnsRecordsResource) Configure(_ context.Context, req resource.Configure
...
 
-	client, ok := req.ProviderData.(*client.Client)
+	pd, ok := req.ProviderData.(*providerData)
 	if !ok {
-		resp.Diagnostics.AddError("Unexpected provider data type", fmt.Sprintf("Expected *client.Client, got %T", req.ProviderData))
+		resp.Diagnostics.AddError("Unexpected provider data type", fmt.Sprintf("Expected *providerData, got %T", req.ProviderData))
 		return
 	}
-	r.client = client
+	r.client = pd.Client
```
kurok, line 108: **Must fix — the plural `dns_records` resource mutates a domain's records but never invalidates the shared cache.**

`providerData` shares one `dnsRecordCache` across every resource, and the cache's own docstring requires that *"any mutation of a domain's records must call `Invalidate(domain)`"*. This resource's `Create`/`Update` (`UpsertDNSRecords`/`DeleteDNSRecords`) and `Delete` (`ClearDNSRecords`) all mutate a domain's records but never invalidate. A later `Find` from the singular `dns_record` resource on the same domain can then be served stale data (e.g. re-upserting a record this resource just deleted) — a divergence the old always-fresh `client.FindDNSRecord` could not produce.

Wire the cache in here:

```suggestion
	r.client = pd.Client
	r.records = pd.DNSRecords
```

Add the field to the resource struct:

```go
type dnsRecordsResource struct {
	client  *client.Client
	records *dnsRecordCache
}
```

And invalidate after each successful write:

```go
// Create / Update — after the Upsert/Delete block succeeds:
r.records.Invalidate(plan.Domain.ValueString())

// Delete — after ClearDNSRecords succeeds:
r.records.Invalidate(state.Domain.ValueString())
```

(If mixing the singular and plural resources on one domain is intentionally unsupported, enforce/document that instead — but these four `Invalidate` calls are cheaper and satisfy the cache's stated invariant.)

internal/provider/dns_record_cache.go
```diff
@@ -0,0 +1,96 @@
...
+		}
+		c.mu.Lock()
+		c.entries[domain] = records
+		c.mu.Unlock()
+		return records, nil
+	})
+	if err != nil {
+		return nil, err
+	}
+	return result.([]client.DNSRecord), nil
```
kurok: **Should fix — `singleflight.Do` binds the shared fetch to the leader caller's `ctx`.**

The fetch closure captures the *first* caller's `ctx`. Terraform refreshes with parallelism 10, so N concurrent reads of one domain collapse into this single flight; if the leader's `ctx` is cancelled (per-op timeout, interrupt), `GetDNSRecords` returns `ctx.Err()` and **every** waiter fails — even those whose own context is still live. This is the well-known singleflight error-sharing footgun.

Use `DoChan` + `select` so each caller observes its own cancellation, and detach the shared fetch from any single caller's ctx so one cancellation can't fail the waiters that still need the result:

```suggestion
	// Cache miss: fetch once across concurrent callers. singleflight collapses
	// the cold-start stampede so a whole refresh wave shares one fetch. DoChan +
	// select lets each caller observe its own context cancellation, and
	// context.WithoutCancel detaches the shared fetch from any single caller's
	// ctx so one cancellation can't fail the waiters that still need the result.
	ch := c.sf.DoChan(domain, func() (any, error) {
		records, err := c.client.GetDNSRecords(context.WithoutCancel(ctx), domain)
		if err != nil {
			return nil, err
		}
		c.mu.Lock()
		c.entries[domain] = records
		c.mu.Unlock()
		return records, nil
	})

	select {
	case <-ctx.Done():
		return nil, ctx.Err()
	case res := <-ch:
		if res.Err != nil {
			return nil, res.Err
		}
		return res.Val.([]client.DNSRecord), nil
	}
```

`context.WithoutCancel` is Go 1.21+ (module is go 1.25). Worth a companion test: concurrent `Find`s where the leader's ctx is cancelled, asserting the others still succeed.

Related lower-priority note: `Invalidate` is just `delete(entries, domain)`, so an `Invalidate` that lands while a fetch is in flight is a no-op and the flight then repopulates stale data. Not reachable today given Terraform's refresh-before-apply phasing, but it's the same invariant as the plural-resource finding above.

internal/provider/dns_records_resource.go
dm1tr1yvovk: Intentional: the singular and plural resources are documented as mutually exclusive per domain. `spaceship_dns_records` owns the entire custom group — its delete+upsert reconciliation would fight any `spaceship_dns_record` on the same zone at the Terraform level regardless of cache freshness, so a mixed config is already broken before stale reads can matter. Rather than defending a configuration we tell users not to write, I tightened the cache docstring in 556aca7: the invariant is now scoped to resources that read through the cache, and the plural resource's non-participation is stated explicitly.

internal/provider/dns_record_cache.go
dm1tr1yvovk: Fixed in 556aca7 — switched to `DoChan` + `select` with `context.WithoutCancel` on the shared fetch, and added the companion test: a cancelled leader fails alone with `context.Canceled` while waiters on the same flight still get the result.

kurok (approved):
Approving — both prior findings are resolved and I found nothing new in the re-review at `556aca7`.

- **singleflight / leader-ctx cancellation** — fixed correctly with `DoChan` + `select` on the caller's ctx and `context.WithoutCancel` on the shared fetch. The detached fetch stays bounded by the 30s `http.Client.Timeout`, so no unbounded background goroutine. The companion test (`TestDNSRecordCache_CancelledCallerDoesNotFailWaiters`) is race-clean and exercises the cancel-leader / live-waiter path.
- **plural `dns_records` cache invalidation** — resolved by design: the docstring now scopes the invariant to cache-reading resources and marks `spaceship_dns_records` as explicitly non-participating. A mixed singular+plural config on one zone already conflicts at the Terraform reconciliation level, so the cache isn't the right place to defend it. Agreed.

Verified locally: `gofmt`/`go vet` clean, provider tests pass under `-race`. LGTM.

PR #92, opened 2026-06-17
ci: add Trivy, govulncheck, SBOM and a consolidated PR security summary

kurok:
## What

Brings this provider's CI in line with [terraform-provider-namecheap#234](https://github.com/namecheap/terraform-provider-namecheap/pull/234): adds the full security-scan stack **and** a single consolidated report posted to every PR.

Surfaced three ways:

- 💬 **Sticky PR comment** — one comment, updated in place per push (marker `<!-- security-scan-summary -->`).
- 🧾 **Job summary** — rendered on the Actions run page.
- 📎 **`security-summary` artifact** — 90-day retention.

## New CI jobs (`.github/workflows/ci.yml`)

| Job | What it does |
|---|---|
| `security` | Trivy fs scan — **vuln + misconfig + secret + license**. Gates on CRITICAL/HIGH *fixable*. License denylist in `trivy.yaml`. Emits CycloneDX SBOM (`sbom-cyclonedx`) + a non-gating JSON pass (`trivy-json`). |
| `govulncheck` | Go-native **reachability** gate (vuln in *called* code fails). Non-gating JSON pass (`govulncheck-json`). |
| `security-report` | Consolidates Trivy + govulncheck + SBOM + open CodeQL alerts into the Markdown report; writes job summary, uploads `security-summary`, posts/updates the sticky comment. |

Report covers: Trivy findings by severity, IaC misconfig, secrets (rule/location only — **never the value**), licenses; govulncheck reachable-vs-imported; open CodeQL alerts (from the existing `codeql.yml`); and a dependency overview (SBOM component count, `go.mod` direct/indirect, license breakdown).

## Design notes

- **Reporting only — never gates.** Per-scan ✅/❌ comes verbatim from the gate jobs' `needs.*.result`, so a green comment can't mask a red gate. The `security` and `govulncheck` jobs are the actual merge gates.
- **PR association:** uses the `pull_request` event number directly (this workflow already triggers on `pull_request`). The comment step is restricted to same-repo branches — **fork PRs skip the comment** (their `GITHUB_TOKEN` is read-only) but still get the job summary + artifact.
- Matches this repo's conventions: 4-space YAML, existing action pins reused (`checkout@v5.0.1`, `setup-go@v6.4.0`, `upload-artifact@v6.0.0`); new `download-artifact` SHA-pinned to `v6.0.0`. Go version via `go-version-file: go.mod`.
- New `trivy.yaml` (license denylist) and `scripts/security-summary.sh` (the generator) are shared verbatim with the namecheap provider. Verified locally: YAML parses, `bash -n`, and the generator against this repo's real `go.mod` (7 direct · 51 indirect) plus fully-degraded inputs.

Documented under **Automated Security Scanning** in `SECURITY.md`.

### DIFF
--- a/scripts/security-summary.sh
+++ b/scripts/security-summary.sh
@@ -1,0 +1,172 @@
+#!/usr/bin/env bash
+# Render a consolidated security-scan summary (GitHub-flavored Markdown) from the
+# JSON outputs of the CI security jobs. Consumed by the `security-report` job in
+# .github/workflows/ci.yml: written to the run's job summary, uploaded as the
+# `security-summary` artifact, and posted as a sticky PR comment.
+#
+# Usage:
+#   security-summary.sh TRIVY_JSON GOVULN_JSON SBOM_JSON GO_MOD CODEQL_JSON
+#
+# Every input is best-effort: a missing, empty, or unparseable file degrades
+# that section gracefully rather than failing. This script is REPORTING ONLY and
+# never exits non-zero on scan content — the merge gates live in the scan jobs.
+# Pass/fail shown in the status table comes from the gate jobs' results
+# (SECURITY_RESULT / GOVULNCHECK_RESULT env), not from re-deriving it here.
+set -uo pipefail
+
+TRIVY="${1:-}"
+GOVULN="${2:-}"
+SBOM="${3:-}"
+GOMOD="${4:-go.mod}"
+CODEQL="${5:-}"
+
+MARKER='<!-- security-scan-summary -->'
+SHA_SHORT="${COMMIT_SHA:0:8}"
+BRANCH="${BRANCH:-unknown}"
+NOW="$(date -u '+%Y-%m-%d %H:%M UTC')"
+
+# jq wrapper: print the filter result, or a fallback when the file is
+# missing/empty/invalid, so one dead scan job can't break the whole report.
+# Args: <file> <filter> <fallback> [slurp-flag e.g. -s]
+jqf() {
+  local file="$1" filter="$2" fallback="$3" slurp="${4:-}"
+  if [ -s "$file" ]; then
+    jq ${slurp} -r "$filter" "$file" 2>/dev/null || printf '%s' "$fallback"
+  else
+    printf '%s' "$fallback"
+  fi
+}
+
+# success -> check, failure -> cross, anything else (skipped/unknown) -> circle
+emoji() {
+  case "$1" in
+    success) printf '✅' ;;
+    failure) printf '❌' ;;
+    *) printf '⚪' ;;
+  esac
+}
+
+# --- Trivy counts ----------------------------------------------------------
+tv_crit=$(jqf "$TRIVY" '[.Results[]?.Vulnerabilities[]? | select(.Severity=="CRITICAL")] | length' 0)
+tv_high=$(jqf "$TRIVY" '[.Results[]?.Vulnerabilities[]? | select(.Severity=="HIGH")]     | length' 0)
+tv_med=$(jqf  "$TRIVY" '[.Results[]?.Vulnerabilities[]? | select(.Severity=="MEDIUM")]   | length' 0)
+tv_low=$(jqf  "$TRIVY" '[.Results[]?.Vulnerabilities[]? | select(.Severity=="LOW")]      | length' 0)
+tv_unkn=$(jqf "$TRIVY" '[.Results[]?.Vulnerabilities[]? | select(.Severity=="UNKNOWN")]  | length' 0)
+tv_misconf=$(jqf "$TRIVY" '[.Results[]?.Misconfigurations[]? | select(.Status=="FAIL")] | length' 0)
+tv_secrets=$(jqf "$TRIVY" '[.Results[]?.Secrets[]?] | length' 0)
+tv_lic=$(jqf "$TRIVY" '[.Results[]?.Licenses[]?] | length' 0)
+
+# --- govulncheck counts (NDJSON stream -> slurp) ---------------------------
+gv_reach=$(jqf "$GOVULN" '[.[] | select(.finding!=null) | .finding | select(.trace[0].function!=null) | .osv] | unique | length' 0 -s)
+gv_total=$(jqf "$GOVULN" '[.[] | select(.finding!=null) | .finding.osv] | unique | length' 0 -s)
+gv_imported=$(( gv_total - gv_reach ))
+[ "$gv_imported" -lt 0 ] && gv_imported=0
+
+# --- SBOM / dependency overview --------------------------------------------
+sbom_comp=$(jqf "$SBOM" '[.components[]?] | length' 0)
+read -r mod_direct mod_indirect < <(awk '
+  /^require[[:space:]]*\(/ { inblk=1; next }
+  inblk && /^\)/ { inblk=0; next }
+  /^require[[:space:]]+[^(]/ { if ($0 ~ /\/\/[[:space:]]*indirect/) ind++; else dir++; next }
+  inblk && NF>=2 && $1!="//" { if ($0 ~ /\/\/[[:space:]]*indirect/) ind++; else dir++ }
+  END { printf "%d %d", dir+0, ind+0 }
+' "$GOMOD" 2>/dev/null || echo "0 0")
+
+# --- CodeQL (code scanning) open alerts ------------------------------------
+cq_total=$(jqf "$CODEQL" 'if type=="array" then length else 0 end' 0)
+cq_crit=$(jqf "$CODEQL"  'if type=="array" then [.[]|select(.rule.security_severity_level=="critical")]|length else 0 end' 0)
+cq_high=$(jqf "$CODEQL"  'if type=="array" then [.[]|select(.rule.security_severity_level=="high")]|length else 0 end' 0)
+
+# --- status from the actual gate jobs --------------------------------------
+trivy_status=$(emoji "${SECURITY_RESULT:-}")
+govuln_status=$(emoji "${GOVULNCHECK_RESULT:-}")
+
+# ===========================================================================
+# Markdown. First line is the sticky marker so the workflow can find + update
+# this exact comment instead of posting a new one each push.
+# ===========================================================================
+printf '%s\n' "$MARKER"
+printf '## 🔐 Security scan summary\n\n'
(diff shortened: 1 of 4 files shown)


scripts/security-summary.sh
dm1tr1yvovk, line 1: maybe add this scrip to makefile?

kurok: It is your discretion, but for the purpose of generating a report, we may need to establish a separate repository for its use in both TF providers.

dm1tr1yvovk:
## Review summary

Thanks for the thorough work here — the report machinery is clearly well-built. My main concern is **scope vs. value for this specific repo**: the PR bundles one genuinely valuable gate (govulncheck) with a lot of scanning whose marginal value for a 7-direct-dep public Terraform provider (no containers, no real IaC, no Dependabot, CodeQL already running) is low-to-zero. I'd like to land the valuable part and shelve the rest.

### What earns its place vs. what doesn't

**Keep — govulncheck.** The clear win. Go-native, call-graph *reachability* (gates only when vulnerable code is actually called), covers stdlib CVEs tied to the toolchain. With no Dependabot here and hand-bumped deps, this fills a real gap with low noise.

**Drop — the whole Trivy `security` job.** Scanner by scanner against what this repo actually contains:
- **`vuln` (SCA):** redundant with govulncheck and weaker — version-range matching, no reachability, more noise on the 51-module graph. The `ignore-unfixed: true` is already an admission of that noise.
- **`misconfig`:** scans IaC, but the repo has no infrastructure — only example `.tf` snippets for `spaceship_dns_records`. Trivy doesn't know the provider's resources; it will find nothing.
- **`secret`:** the repo is public, so GitHub native secret scanning + push protection are free and *better* (block at push time, not after it's in history). Trivy adds a weaker second detector prone to flagging example key-shaped values.
- **`license` denylist:** guards against GPL/AGPL/SSPL in a tree that's entirely HashiCorp (MPL/MIT), golang.org/x (BSD), protobuf/testify. Realistic risk ≈ nil, and it adds `trivy.yaml` + an exception-PR process for a problem that won't occur.

**Drop — the SBOM.** It's for consumers of a distributed binary; this ships via the Terraform Registry and nobody ingests a per-PR SBOM. If wanted, generate it once at *release* time (GoReleaser supports it), not every push.

**Drop — the `security-report` job + 172-line script + sticky PR comment.** Biggest maintenance cost, lowest value: once Trivy is gone its sections render zeros; a comment updated on every push is noise next to the Checks UI; 172 lines of untested `jq`/`awk` where a typo silently degrades to "0 findings" (the worst failure mode for a security summary); and it's copy-pasted from the namecheap provider rather than a shared action, so it will drift.

This would take the PR from ~410 lines / 3 jobs / 1 script / 1 config down to roughly a 15-line single govulncheck job.

### Correctness bugs (if any of it ships as-is)

1. **CodeQL alerts always report 0 on PRs.** The "Fetch open CodeQL alerts" step queries `?ref=${GITHUB_REF}`, but on `pull_request` events `GITHUB_REF` is `refs/pull/N/merge` — a ref CodeQL never analyzes. The call silently returns `[]`, so the summary always shows "CodeQL: 0 open" even with real open alerts. Should use `github.head_ref`.
2. **`paths-ignore` vs. required checks = merge deadlock.** The workflow-level `paths-ignore` skips the entire run for docs-only PRs. If `security`/`govulncheck` are ever made required checks, a docs-only PR leaves the required check "Expected — waiting for status" indefinitely.
3. **Markdown injection.** Scan values (`PkgName`, `Title`) and branch names are interpolated raw into table cells; a value containing `|` or a backtick corrupts the rendered table.
4. **`SHA_SHORT="${COMMIT_SHA:0:8}"`** has no empty-guard while `BRANCH` gets `:-unknown` — inconsistent; an empty SHA renders empty backticks.
5. **Trivy runs 3× in one job**, each pulling the vuln DB fresh — wasted time and ghcr rate-limit exposure (moot if Trivy is dropped).

The govulncheck NDJSON parsing, the `go.mod` awk parser, `merge-multiple` artifact flattening, and the fork-PR comment skip all looked correct.

### Suggestion

Could we **split this**? Land govulncheck as a small gating job (dropping its JSON/artifact steps, which only fed the removed report), and shelve Trivy + SBOM + the report machinery. If cross-repo consolidated reporting is a real goal, it'd be better as a versioned reusable workflow than a copy-pasted script — but that's a bigger discussion than this repo's risk surface needs.

kurok:
Thanks for the depth here — this is a genuinely useful review. I've pushed fixes for the two real correctness bugs, and want to respond on scope and the remaining three findings, which I verified against the actual toolchain rather than applying the suggested changes blind.

## Fixed in `ce40b79` (`scripts/security-summary.sh`)

- **#3 — Markdown injection.** Added a reusable jq `cell` helper that, for every untrusted field interpolated into a table cell, escapes `|` → `\|`, collapses CR/LF to spaces, and strips backticks so a value can't break the table or forge/escape an inline-code span. Applied across all row builders (vulnerabilities, reachable govulncheck — including the back-ticked symbol, license breakdown, license findings, IaC misconfig, secrets). The shell-side branch name is sanitized the same way before it lands in its backtick span. Secrets still print rule/severity/title/location only — never the value. Verified with crafted inputs containing `|`, backticks, and newlines.
- **#4 — Empty/unset SHA.** Now `_sha="${COMMIT_SHA:-unknown}"; SHA_SHORT="${_sha:0:8}"`, consistent with `BRANCH`. Guarding before the substring also avoids the `set -u` unbound-variable abort when `COMMIT_SHA` is entirely unset. Empty and unset both render `unknown`; a real SHA still renders its first 8 chars.

## On the "drop it to ~15 lines" scope critique

You're right on several points and I want to concede them up front: for a provider with this few direct dependencies, the *marginal* detection value of some Trivy scanners (IaC misconfig, secret) is low; a copy-pasted summary script will drift between the two providers; and the jq in `security-summary.sh` isn't unit-tested. Those are fair.

I'm still keeping the full scope, for three reasons. First, parity: `terraform-provider-namecheap#234` lands the same Trivy + govulncheck + SBOM + consolidated-summary shape, and a consistent, predictable security posture across both public TF providers is worth more than per-repo micro-optimization — a reviewer or downstream consumer should see the same gates and the same report in both. Second, defense-in-depth: govulncheck (reachability) and Trivy SCA (version-range) catch different classes, and license-policy gating plus an SBOM artifact are registry-distribution requirements, not noise. Third — and this is the actual product goal — the single sticky PR comment is the deliverable: one always-current place to read every gate's result, rather than asking people to open four artifacts. Stripping to bare govulncheck would delete that, which is the feature.

On the strongest criticism (drift + untested script + the inline "add to a makefile?" thread): I'd rather fix the root cause than trim. Concrete follow-up — extract the report generator into a **shared, versioned reusable workflow** (`workflow_call`) consumed by both providers, with the jq logic moved behind a small tested entrypoint. That removes the copy-paste entirely, gives the script a single place to be tested, and answers the makefile question by making the generator invokable locally too. I'll file that as its own PR rather than expand this one.

## The other three correctness findings

**#1 — CodeQL `?ref=${GITHUB_REF}`: no change (verified correct as-is).** The code-scanning REST API documents `refs/pull/<number>/merge` as the canonical PR ref, and PR alert instances are recorded under exactly that ref. `codeql.yml` runs on `pull_request` (advanced setup), so CodeQL analyzes the *merge* commit and files alerts under `refs/pull/N/merge`. The `security-report` job is also `pull_request`-triggered, so `GITHUB_REF` is already `refs/pull/N/merge` — the right ref. `github.head_ref` would query `refs/heads/<feature-branch>`, which CodeQL never analyzes here (no feature-branch push trigger), making the count *always* zero — the opposite of the intent. The one genuine edge case is the race where CodeQL hasn't finished analyzing the PR yet; that already degrades to `[]`, which the step comment flags as best-effort.

**#2 — `paths-ignore` vs. required-check deadlock: real, but pre-dates this PR and is out of scope.** The top-level `paths-ignore` block already exists at the base commit (`dd693c9`) and is mirrored in `codeql.yml`; it currently governs the existing required `check` job and `acceptance`. This PR didn't add it, and the new `security`/`govulncheck` jobs aren't required checks, so no deadlock exists today. The clean fixes (a docs-only "skip" companion satisfying the required-check name, or moving from workflow-level `paths-ignore` to per-job path filters) all change the behavior of the pre-existing `check`/`acceptance` jobs and `codeql.yml` — a repo-wide change. I'll address it consistently across `ci.yml` and `codeql.yml` in a separate PR at the point those checks are marked required, rather than touch unrelated jobs here.

**#5 — Trivy runs 3×: premise is partly off, and the suggested collapse isn't implementable on the pinned version.** With `cache: "true"` the action stores the vuln/Java DB + checks bundle under `$GITHUB_WORKSPACE/.cache/trivy` and restores it before each scan, so steps 2 and 3 reuse the local cache instead of re-downloading — that already covers the ghcr rate-limit concern. The `trivy convert` collapse fails on pinned v0.71.1: `trivy convert --severity CRITICAL,HIGH --ignore-unfixed --exit-code 1 … <json>` → `FATAL … unknown flag: --ignore-unfixed` (convert has no such flag or config key — only the `fs` scan config does). Since the gate is intentionally *fixable-only*, re-deriving it without that filter would regress it into failing on unfixed CVEs. The gating pass and the all-severity JSON also can't be derived from each other without losing data, so I kept both passes. Happy to revisit if a future Trivy adds `--ignore-unfixed` to `convert`.

dm1tr1yvovk (approved):
Approving. The two real correctness bugs (#3 Markdown injection, #4 SHA guard) are fixed cleanly — generalized into the reusable `cell` helper rather than patched per-site, which is the right depth. And you're correct on the two I got wrong: #1 (`refs/pull/N/merge` is the canonical PR ref for PR-triggered CodeQL, so `GITHUB_REF` is right and `head_ref` would always read zero) and #5 (steps share the workspace filesystem, so no DB re-download; and `convert` lacks `--ignore-unfixed` on v0.71.1). #2 is real but pre-existing and correctly deferred to when these become required checks.

On scope: keeping the full Trivy + SBOM + summary shape for org-wide parity is a reasonable maintainer call, and it's easily reversible if the per-push comment proves noisy. The one thing I'd like to hold us to is the reusable `workflow_call` follow-up you proposed — that's the actual fix for the copy-paste-drift + untested-jq concern, which is the only durable cost here. Let's track it as a separate issue so it doesn't slip.

Nice work verifying against the real toolchain instead of applying suggestions blind.

PR #96, opened 2026-06-25
fix(client): paginate domain list

dm1tr1yvovk:
## Description

Prepares the Spaceship API client for extraction into a standalone SDK (`go-spaceship-sdk`) by polishing it in place first. Three focused, independently-building commits:

- **`fix(client): paginate domain list`** — `GetDomainList` previously fetched only the first 100 domains (`take=100, skip=0`, no loop), silently truncating larger accounts. `GetDomainInfo`'s 429 rate-limit fallback relies on this list and inherited the same gap. It now follows pagination until all pages are retrieved, mirroring the existing `GetDNSRecords` loop. An unexported `listDomains(ctx, pageSize)` seam lets tests force multi-page behavior without owning 100 domains.
- **`refactor(client): correct CustomNamerverProvider typo`** — renames the exported const to `CustomNameserverProvider` while the package is still `internal/`, so the misspelling never becomes a public API contract once the SDK is extracted.
- **`docs(client): document transport and domain methods`** — adds godoc to the request/error/domain methods and types, fixes comments to begin with the identifier name (revive convention), and promotes the 429 fallback from an informal inline comment to a proper doc note. No behavior change.

### Verification of the pagination fix

- Unit test `TestGetDomainList_Paginates` proves the multi-page loop against a mock server.
- The `take`/`skip` params, `{items, total}` response shape, `orderBy=name`, and the **100-item page cap** were verified against the official API docs (docs.spaceship.dev).
- A credential-gated live test (`TestAccGetDomainListPagination`) confirms the contract against the **real API**: it compares `take=1` vs `take=100` retrieval and asserts the `total == len(items)` termination invariant.

> **Note on coverage:** the test account currently holds a single domain, so the multi-page *iteration* branch is not exercised live yet. The test detects this and logs it explicitly rather than passing silently, and auto-upgrades to full multi-page proof once it runs against an account with ≥ 2 domains.

## Related Issues

N/A

## Checklist

- [x] Unit tests added or updated
- [x] `make test` passes
- [x] `golangci-lint run ./...` passes
- [x] `go build .` passes
- [x] `make docs` run if schema changed (and changes committed) — no schema change (client-only), N/A

## Acceptance Test Output

```console
=== RUN   TestAccGetDomainListPagination
    domains_acc_test.go:53: contract confirmed (take/skip/total/orderBy) but only 1 domain in account; multi-page iteration NOT exercised — needs >= 2 domains
--- PASS: TestAccGetDomainListPagination (0.55s)
PASS
ok      terraform-provider-spaceship/internal/client    1.389s
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/client/domains.go
+++ b/internal/client/domains.go
@@ -7,7 +7,8 @@ import (
 	"strconv"
 )
 
-// DomainList is a page of domains returned by the domain list endpoint.
+// DomainList holds domains returned by the domain list endpoint. GetDomainList
+// returns one with every domain in the account, accumulated across all pages.
 type DomainList struct {
 	Items []DomainInfo `json:"items"`
 	Total int64        `json:"total"`
--- a/internal/client/domains.go
+++ b/internal/client/domains.go
@@ -7,10 +7,14 @@ import (
 	"strconv"
 )
 
+// DomainList is a page of domains returned by the domain list endpoint.
 type DomainList struct {
 	Items []DomainInfo `json:"items"`
 	Total int64        `json:"total"`
 }
+
+// DomainInfo describes a single domain: registration state, lifecycle and
+// verification status, privacy settings, nameservers, and contacts.
 type DomainInfo struct {
 	Name               string            `json:"name"`
 	UnicodeName        string            `json:"unicodeName"`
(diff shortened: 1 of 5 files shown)


dm1tr1yvovk:
On the `TestAccGetDomainListPagination` placement: it lives at the client layer and hits the live API rather than exercising a Terraform lifecycle, which differs from the CLAUDE.md testing-strategy convention. That is intentional here.

Its unique value is asserting that the real API's `total` is a **global** count, not a per-page one. The pagination loop's termination depends on that assumption (`len(items) >= total || empty page`), and a mock-based unit test cannot prove it — the mock just returns whatever `total` we tell it to. Only a live call verifies the contract holds.

This client is being prepared for extraction into a standalone SDK, at which point the provider-level "acc tests = Terraform lifecycle" rule no longer governs it. Rather than reshape the test now only to move it to its own repo shortly, leaving it as a client-layer integration test is the lower-churn choice.

PR #100, opened 2026-07-03
fix(records): reject apex ALIAS records at plan time

dm1tr1yvovk:
## What

Reject apex ALIAS records (`type = "ALIAS"`, `name = "@"`) at plan time.

## Why

Spaceship silently stores an apex ALIAS as a **root CNAME**, not an ALIAS. Because the provider matches records by type + name + data, the read-back CNAME never matches the declared ALIAS, so the record is recreated on **every apply** (non-convergence).

The validator now rejects `name = "@"` for ALIAS with a clear diagnostic pointing users at the form the API actually keeps:

```hcl
type = "CNAME", name = "@", cname = "<target>"
```

This is distinct from the existing `alias_name = "@"` rejection (the *target* field), which is enforced by the SDK's `ValidateAliasName`.

## Tests

- Unit: apex rejected, non-apex allowed, non-ALIAS types ignored.
- Acc (single resource): `TestAccDNSRecord_aliasApexNameFailsPlan`.
- Acc (multiple-record resource): apex alone, and apex mixed among valid records.

`make test`, `make lint`, and `go build .` all pass; acceptance tests confirmed passing locally.

### DIFF
--- a/internal/docs/dns-records.md
+++ b/internal/docs/dns-records.md
@@ -29,6 +29,8 @@ The provider's `recordValueSignature()` function follows the same rules. All fie
 ### Record type notes
 
 - **ALIAS**: Resolves a canonical (true) domain name. Implements CNAME-like behavior for the zone apex where CNAME is not allowed. The `aliasName` field is a hostname (1-253 chars, `hostNameValue` pattern).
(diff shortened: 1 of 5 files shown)


dm1tr1yvovk:
## Review summary — Approve ✅

Clean, correctly-layered fix. The apex-ALIAS rejection is a **reconciliation guard** (not an API rule): the API accepts apex ALIAS but stores it as a root CNAME, so type+name+data matching never re-matches → perpetual recreate. Rejecting at plan time with a diagnostic pointing at the `type = "CNAME", name = "@"` form is the right layer.

**Strengths**
- Check is scoped after the `type == "ALIAS"` short-circuit, so `@` stays legal for other record types.
- Defensive guard (`!IsNull && !IsUnknown && == "@"`) lets unknown/null names fall through correctly during early plan phases.
- Existing `alias_name = "@"` *target* rejection left intact; the test comment was updated to disambiguate the two cases.
- Well-layered tests (unit in the SDK-free package + acc tests covering apex-alone and apex-among-valid-records); all pass locally.

**Minor (non-blocking)**
- Apex is matched only as `"@"` — worth confirming an empty `name` isn't also an accepted apex spelling.
- The validator comment still says name/ttl are handled by schema-level validators; a one-liner noting this ALIAS-specific name check would avoid a future reader seeing a contradiction.

Low risk overall — purely additive validation that can only reject already-broken (perpetual-recreate) configs.

PR #114, opened 2026-07-08
fix(domain): apply configured auto_renew and nameservers on create

dm1tr1yvovk:
## Problem

The domain resource adopts existing domains on create, but `Create` stored whatever `GetDomainInfo` returned while ignoring the planned `auto_renew` and `nameservers`. Any config/infra mismatch on the first apply made Terraform fail with:

```
Error: Provider produced inconsistent result after apply
.auto_renew: was cty.True, but now cty.False.
```

Since the error fires on the first apply, `Update` (which handles both settings correctly) was never reachable — `auto_renew` could effectively never be changed.

## Fix

`Create` now mirrors `Update`'s converge-then-trust pattern:

- If the configured `auto_renew` differs from the domain's actual value, push it via `UpdateAutoRenew` (no write when they already match).
- Same for `nameservers` via a new shared `pushNameservers` helper, now used by both `Create` and `Update`. The converge is skipped when nested attributes are still unknown at create (no prior state for `UseStateForUnknown` to resolve), falling back to plain adoption.
- State trusts the plan values over the pre-update read, same stale-read rule as `Update`.

## Tests

New acceptance tests force the real domain to the opposite setting via the SDK in `PreConfig` before the first apply, cover the reverse direction through a tainted recreate, and assert an empty re-plan:

- `TestAccDomain_autoRenewalMismatchOnCreate`
- `TestAccDomain_nameserversMismatchOnCreate`

Both failed with the exact reported error before the fix and pass after. Existing `TestAccDomain_autoRenewal`, `TestAccDomain_nameservers`, and `TestAccDomain_nameserversValidationErrors` still pass against the real API.

### DIFF
--- a/internal/provider/domain_resource.go
+++ b/internal/provider/domain_resource.go
@@ -295,11 +295,56 @@ func (d *domainResource) Create(ctx context.Context, req resource.CreateRequest,
 		return
 	}
 
-	domainInfo, err := d.client.GetDomainInfo(ctx, plan.Domain.ValueString())
+	domainName := plan.Domain.ValueString()
+
+	domainInfo, err := d.client.GetDomainInfo(ctx, domainName)
 
 	if err != nil {
 		resp.Diagnostics.AddError("Unable to read domain info", err.Error())
 		return
+	}
+
+	// Adoption must converge the domain to the configured settings. Otherwise
+	// a config/infra mismatch on the first apply stores the actual values
+	// while the plan promised the configured ones, and Terraform fails with
+	// "Provider produced inconsistent result after apply".
+	if !plan.AutoRenew.IsNull() && !plan.AutoRenew.IsUnknown() && plan.AutoRenew.ValueBool() != domainInfo.AutoRenew {
+		_, err := d.client.UpdateAutoRenew(ctx, domainName, plan.AutoRenew.ValueBool())
+		if err != nil {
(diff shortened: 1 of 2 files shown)


dm1tr1yvovk:
## Independent review summary

**Scope reviewed:** the full diff (`4dd3d6d..7591d52`) — Create-path convergence for `auto_renew` and `nameservers`, the shared `pushNameservers` helper, and both mismatch-on-create acceptance tests. Every known/unknown plan-value combination was traced through Create, including the ModifyPlan interaction and unknown nested attributes; `Update`, `Read`, and `Import` were verified unchanged.

**Result:** ready to merge — no critical issues.
- No path writes unknown values to state, skips a needed PUT, or issues a redundant one; no-op first applies perform zero writes.
- One review finding already addressed in `7591d52`: both converge directions are now pinned by refresh-backed `ExpectEmptyPlan` steps instead of state-only checks.
- One pre-existing gap noted (not a regression, out of scope): `nameservers` with `hosts` but no `provider` still fails the first apply with the same error, since `provider` is unknown at create. Suggested follow-up: require `provider` in the validator when the block is present.

**Verification:** new acceptance tests failed with the exact reported error before the fix and pass after, against the real API; existing domain acceptance tests, `make test`, `make lint`, and build are all green.

PR #122, opened 2026-07-23
feat: retry rate-limited API calls with configurable operation timeouts

dm1tr1yvovk:
## Problem

The provider did not retry on HTTP 429. Several Spaceship endpoints are aggressively rate limited per domain, so ordinary plan → apply → re-plan workflows failed intermittently with raw \`spaceship api error (status 429)\` errors.

## Approach

- The SDK (v0.2.0) now exposes the 429 \`Retry-After\` header (delta-seconds) as \`SpaceshipApiError.RetryAfter\` plus an \`IsRateLimitError\` helper — facts only, the client never sleeps.
- A provider-side \`withRetry\` helper owns the policy: it retries only 429s, sleeps exactly the server-requested wait (30s fallback + 1s margin), logs each wait via \`tflog.Warn\`, aborts immediately on Ctrl-C, and fails fast with an actionable message when the requested wait cannot fit the remaining deadline.
- Every resource and data source gains a \`timeouts\` block whose deadline is the retry loop's only budget. Defaults are derived from rate-limitable calls per operation × one full throttling window, and are overridable per resource.

## Coverage

- \`spaceship_domain\` (create/read/update) and both domain data sources (read)
- \`spaceship_personal_nameserver\` (all four operations)
- \`spaceship_dns_records\` and \`spaceship_dns_record\` (all four operations; shared retry wrappers in \`dns_record_common.go\`, singular resource retries around the shared cache so waits respect each caller's deadline)

The SDK's existing domain-list fallback for \`GetDomainInfo\` is untouched and still runs before any waiting happens.

## Testing

- \`withRetry\` unit tests: retry-then-success, non-429 passthrough, missing-header default, fail-fast on unfittable wait, prompt cancellation mid-sleep (injected sleep, no real waits)
- End-to-end 429 wiring test: mock returns 429 + \`Retry-After\` with the list fallback answered empty; asserts the read is retried after the exact requested wait
- Schema tests for every new \`timeouts\` block; existing lifecycle tests unchanged and green
- Design notes in \`internal/docs/rate-limits.md\`


internal/provider/dns_record_resource.go
```diff
@@ -272,7 +338,13 @@ func (r *dnsRecordResource) Update(ctx context.Context, req resource.UpdateReque
...
 	}
 
-	record, err := r.records.Find(ctx, domain, recordType, name, signature)
+	ctx, cancel := operationContext(ctx, plan.Timeouts.Update, dnsRecordUpdateTimeout, &resp.Diagnostics)
+	defer cancel()
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	record, err := r.findRecordWithRetry(ctx, domain, recordType, name, signature)
```
kurok, line 355: The schema change makes the comment above (L322–324) stale in a way that bites: the `timeouts` block doesn't require replacement, so a timeouts-only config edit — including just *adding* the block to existing resources, which is exactly the adoption path for this PR — now plans an in-place update. Update then unconditionally does the zone fetch + a real PUT, and every save invalidates the shared zone cache, so with N `spaceship_dns_record` resources on one domain a timeouts-only apply can burn up to ~2N rate-limited calls against the per-domain bucket, queue later records behind multiple throttling windows, and even blow their update deadline and fail the apply.

Since `ttl` is the only other in-place attribute, an unchanged TTL means there's nothing to write:

```suggestion
	// The timeouts block is client-side only: if ttl is unchanged there is
	// nothing to write, so skip the lookup and upsert entirely.
	if plan.TTL.Equal(state.TTL) {
		plan.ID = state.ID
		resp.Diagnostics.Append(resp.State.Set(ctx, &plan)...)
		return
	}

	ctx, cancel := operationContext(ctx, plan.Timeouts.Update, dnsRecordUpdateTimeout, &resp.Diagnostics)
	defer cancel()
	if resp.Diagnostics.HasError() {
		return
	}

	record, err := r.findRecordWithRetry(ctx, domain, recordType, name, signature)
```

(`plan.TTL` is always known here — optional+computed with a static default — and `plan.ID` carries state's value via `UseStateForUnknown`, so the early return keeps plan consistency. Worth also rewording L322–324 to mention the timeouts block.)

dm1tr1yvovk: Applied in 1789bee — early return when `ttl` is unchanged (before the timeout context spins up), and reworded the stale comment above. Verified live with a dev-override build: adding just the `timeouts` block to an existing record applies in-place with no zone fetch/upsert, while a real `ttl` change still does the roundtrip.

kurok (approved):
Re-checked at 1789bee — the guard lands exactly as discussed: early return before the timeout context spins up, plan consistency holds (ID carried over, new timeouts values reach state), and a real ttl drift still flows through to the write. gofmt/vet and the unit suite are clean locally at this head, CI is green. Optional follow-up, non-blocking: a small unit test asserting a timeouts-only update makes zero API calls would pin the guard against future regressions — the live dev-override verification covers it for now.

PR #138, opened 2026-09-09
fix(deps): bump Go toolchain and vulnerable transitive deps

dm1tr1yvovk:
Both CI security gates are failing on `master` and on every open PR. The last green run on `master` was 2026-08-13; these advisories were published since, so nothing in the code caused this — the scanners simply re-ran against a stale dependency set.

## govulncheck — 4 reachable stdlib advisories

All four are attributed to the `toolchain go1.26.5` directive, not to any dependency, and all are fixed in `go1.26.6`:

| Advisory | Package |
|---|---|
| GO-2026-6218 | `net/url` |
| GO-2026-6090 | `crypto/tls` |
| GO-2026-5972 | `encoding/asn1` |
| GO-2026-5026 | `net/http` |

## Trivy — 5 HIGH/CRITICAL in indirect deps

The gate runs with `ignore-unfixed: true`, so every finding has a patched version available.

| Dependency | Was | Now | CVE | Severity |
|---|---|---|---|---|
| `golang.org/x/crypto` | v0.53.0 | v0.57.0 | CVE-2026-56854 (`ssh`) | CRITICAL |
| `golang.org/x/mod` | v0.37.0 | v0.41.0 | CVE-2026-56864, CVE-2026-56865 (`sumdb`) | HIGH |
| `google.golang.org/grpc` | v1.82.1 | v1.83.2 | CVE-2026-84304, CVE-2026-84445 | HIGH |

CVE-2026-84445 is only fixed in grpc 1.83.2, so 1.83.1 would not have been enough.

Every HashiCorp direct dependency is already at its latest release, so bumping a parent could not pull the fixed transitives — these have to be explicit `// indirect` pins until upstream catches up.

## Note on the `go` directive

`go get` raised the `go` directive from 1.25.9 to 1.26.0. This is forced by `golang.org/x/sync@v0.23.0` ("all: upgrade go directive to at least 1.26.0"). Registry users consume a prebuilt binary, so the floor only affects building from source.

## Supersedes #137

The bump pulled in `golang.org/x/sync` v0.23.0 on its own, so this contains #137's change in full. Dependabot should close that PR automatically once this merges.

## Verification

Run locally against the CI-pinned scanner versions (govulncheck v1.3.0, Trivy v0.71.1 with `trivy.yaml`):

```
govulncheck ./...   exit 0   "No vulnerabilities found."
trivy fs .          exit 0   go.mod: 0 vulnerabilities, 0 license findings
make test           ok       provider 7.1s, records 1.2s
make lint           0 issues
go build .          ok
```


dm1tr1yvovk:
Superseded — folded into #137 as its own commit so the whole change ships as one PR. Same go.mod/go.sum tree (verified byte-identical).

